PR #5, opened 2023-05-24
feat(editor-document): implement addBlock, removeBlock, getBlock methods

ilyamore88:
I've fixed the configuration of Eslint to support Jest. Also, I've implemented new methods for EditorDocument and added unit tests for them.

### DIFF
--- a/src/entities/EditorDocument/EditorDocument.spec.ts
+++ b/src/entities/EditorDocument/EditorDocument.spec.ts
@@ -211,4 +211,135 @@ describe('EditorDocument', () => {
       });
     });
   });
+
+  describe('removeBlock', () => {
+    let document: EditorDocument;
+    let blocksInDocument = 0;
+
+    beforeEach(() => {
+      document = createDocument();
+
+      const block1 = createBlock({
+        name: createBlockNodeName('block1'),
+        parent: document,
+      });
+      const block2 = createBlock({
+        name: createBlockNodeName('block2'),
+        parent: document,
+      });
+      const block3 = createBlock({
+        name: createBlockNodeName('block3'),
+        parent: document,
+      });
+
+      document.addBlock(block1);
+      document.addBlock(block2);
+      document.addBlock(block3);
+
+      // eslint-disable-next-line @typescript-eslint/no-magic-numbers
+      blocksInDocument = 3;
+    });
+
+    it('should remove the block from the beginning of the document', () => {
+      // Arrange
+      const block = document.getBlock(0);
+
+      // Act
+      document.removeBlock(0);
+
+      // Assert
+      expect(document.getBlock(0)).not.toBe(block);
+    });
+
+    it('should remove the block from the specified index', () => {
+      // Arrange
+      const block = document.getBlock(1);
+
+      // Act
+      document.removeBlock(1);
+
+      // Assert
+      expect(document.getBlock(1)).not.toBe(block);
+    });
+
+    it('should remove the block from the end of the document', () => {
+      // Act
+      document.removeBlock(blocksInDocument - 1);
+
+      // Assert
+      expect(() => document.getBlock(blocksInDocument - 1)).toThrowError('Index out of bounds');
+    });
+
+    it('should throw an error if index is greater then block nodes length', () => {
+      // Act
+      const action = (): void | never => document.removeBlock(blocksInDocument);
+
+      // Assert
+      expect(action).toThrowError('Index out of bounds');
+    });
+
+    it('should throw an error if index is less then 0', () => {
+      // Act
+      const action = (): void | never => document.removeBlock(-1);
+
+      // Assert
+      expect(action).toThrowError('Index out of bounds');
+    });
+  });
+
+  describe('getBlock', () => {
+    let document: EditorDocument;
+    let blocksInDocument = 0;
+
+    beforeEach(() => {
+      document = createDocument();
+
+      const block1 = createBlock({
+        name: createBlockNodeName('block1'),
+        parent: document,
+      });
+      const block2 = createBlock({
+        name: createBlockNodeName('block2'),
+        parent: document,
+      });
+      const block3 = createBlock({
+        name: createBlockNodeName('block3'),
+        parent: document,
+      });
+
+      document.addBlock(block1);
+      document.addBlock(block2);
+      document.addBlock(block3);
+
+      // eslint-disable-next-line @typescript-eslint/no-magic-numbers
+      blocksInDocument = 3;
+    });
+
+    it('should return the block from the specific index', () => {
+      // Arrange
+      const index = 1;
+
+      // Act
+      const block = document.getBlock(index);
+
+      // Assert
+      expect(block).toBe(document.getBlock(index));
+    });
+
+    it('should throw an error if index is greater then block nodes length', () => {
+      // Act
+      const action = (): BlockNode | never => document.getBlock(blocksInDocument);
+
+      // Assert
+      expect(action).toThrowError('Index out of bounds');
+    });
+
+    it('should throw an error if index is less then 0', () => {
+      // Act
+      const action = (): BlockNode | never => document.getBlock(-1);
+
+      // Assert
+      expect(action).toThrowError('Index out of bounds');
+    });
+  });
 });
--- a/src/entities/EditorDocument/EditorDocument.spec.ts
+++ b/src/entities/EditorDocument/EditorDocument.spec.ts
@@ -126,7 +126,7 @@ describe('EditorDocument', () => {
       });
 
       // Act
-      const action = (): void | never => document.addBlock(block, document.length + 1);
+      const action = (): void => document.addBlock(block, document.length + 1);
 
       // Assert
       expect(action).toThrowError('Index out of bounds');
@@ -139,7 +139,7 @@ describe('EditorDocument', () => {
       });
 
       // Act
-      const action = (): void | never => document.addBlock(block, -1);
+      const action = (): void => document.addBlock(block, -1);
 
       // Assert
       expect(action).toThrowError('Index out of bounds');
@@ -182,7 +182,7 @@ describe('EditorDocument', () => {
 
     it('should throw an error if index is greater then block nodes length', () => {
       // Act
-      const action = (): void | never => document.removeBlock(document.length);
+      const action = (): void => document.removeBlock(document.length);
 
       // Assert
       expect(action).toThrowError('Index out of bounds');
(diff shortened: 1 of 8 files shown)


src/entities/EditorDocument/index.ts
```diff
@@ -27,4 +27,57 @@ export class EditorDocument {
...
+  }
+
+  /**
+   * Removes a BlockNode from the EditorDocument at the specified index.
+   * Throws an error if the index is out of bounds.
+   *
+   * @param index - The index of the BlockNode to remove
+   */
+  public removeBlock(index: number): void | never {
+    if (index < 0 || index >= this.#children.length) {
```
gohabereg, line 63: You can extract that to a separate method to use in both addBlock and removeBlock

src/entities/EditorDocument/EditorDocument.spec.ts
```diff
@@ -0,0 +1,203 @@
...
+      name: createBlockNodeName('block3'),
+      parent: document,
+    });
+
+    document.addBlock(block1);
+    document.addBlock(block2);
+    document.addBlock(block3);
+
+    // eslint-disable-next-line @typescript-eslint/no-magic-numbers -- 3 is the number of blocks in the document before each test
+    blocksInDocument = 3;
```
gohabereg, line 42: You could just implement `length` getter for document class

src/entities/EditorDocument/EditorDocument.spec.ts
```diff
@@ -0,0 +1,203 @@
...
+      document.addBlock(block);
+
+      // Assert
+      expect(document.getBlock(blocksInDocument)).toBe(block);
+    });
+
+    it('should add the block to the beginning of the document if index is 0', () => {
+      // Arrange
+      const block = createBlock({
+        name: createBlockNodeName('block'),
```
gohabereg, line 62: Unnecessary line

src/entities/EditorDocument/EditorDocument.spec.ts
```diff
@@ -0,0 +1,203 @@
...
+      // Act
+      const action = (): void | never => document.addBlock(block, -1);
+
+      // Assert
+      expect(action).toThrowError('Index out of bounds');
+    });
+  });
+
+  describe('removeBlock', () => {
+    it('should remove the block from the beginning of the document', () => {
```
gohabereg, line 128: If you add length getter, you can also check it here to be sure block is removed

src/entities/EditorDocument/EditorDocument.spec.ts
```diff
@@ -0,0 +1,203 @@
...
+  describe('getBlock', () => {
+    it('should return the block from the specific index', () => {
+      // Arrange
+      const index = 1;
+
+      // Act
+      const block = document.getBlock(index);
+
+      // Assert
+      expect(block).toBe(document.getBlock(index));
```
gohabereg, line 184: This test doesn't make any sense. You need to get expected value from somewhere else (from beforeEach hook for example)

gohabereg (changes requested):

src/entities/EditorDocument/index.ts
```diff
@@ -27,4 +27,73 @@ export class EditorDocument {
...
+    if (index === undefined) {
+      this.#children.push(blockNode);
+
+      return;
+    }
+
+    /**
+     * Throws error if index is out of bounds
+     */
+    if (index < 0 || index > this.#children.length) {
```
gohabereg, line 56: You can use `isIndexOutOfBounds` here as well

src/entities/EditorDocument/index.ts
```diff
@@ -27,4 +27,73 @@ export class EditorDocument {
...
+
+  /**
+   * Removes a BlockNode from the EditorDocument at the specified index.
+   * Throws an error if the index is out of bounds.
+   *
+   * @param index - The index of the BlockNode to remove
+   */
+  public removeBlock(index: number): void | never {
+    if (this.isIndexOutOfBounds(index)) {
+      throw new Error('Index out of bounds');
```
gohabereg, line 71: You can throw an error inside `isIndexOutOfBounds`

gohabereg (changes requested):

gohabereg (approved):

src/entities/EditorDocument/EditorDocument.spec.ts
```diff
@@ -0,0 +1,228 @@
...
+
+    for (let i = 0; i < countOfBlocks; i++) {
+      const block = createBlock({
+        parent: document,
+      });
+
+      document.addBlock(block);
+      blocks.push(block);
+    }
+  });
```
neSpecc, line 36: lets create a document inside cases instead of beforeEach. Because "arrange" could have differences

src/entities/EditorDocument/EditorDocument.spec.ts
```diff
@@ -0,0 +1,228 @@
...
+
+  describe('length', () => {
+    it('should return the number of blocks in the document', () => {
+      // Arrange
+      const expected = 3;
+      const document1 = new EditorDocument({
```
neSpecc, line 42: bad naming: `document`, `document1`, `document2`

src/entities/EditorDocument/EditorDocument.spec.ts
```diff
@@ -0,0 +1,228 @@
...
+
+      document.addBlock(block);
+      blocks.push(block);
+    }
+  });
+
+  describe('length', () => {
+    it('should return the number of blocks in the document', () => {
+      // Arrange
+      const expected = 3;
```
neSpecc, line 41:
```suggestion
      const blocksCount = 3;
```

src/entities/EditorDocument/EditorDocument.spec.ts
```diff
@@ -0,0 +1,228 @@
...
+
+      // Act
+      const actual = document1.length;
+
+      // Assert
+      expect(actual).toBe(expected);
+    });
+  });
+
+  describe('addBlock', () => {
```
neSpecc, line 65:
```suggestion
  describe('addBlock()', () => {
```

src/entities/EditorDocument/EditorDocument.spec.ts
```diff
@@ -0,0 +1,228 @@
...
+      });
+
+      // Act
+      document.addBlock(block, 1);
+
+      // Assert
+      expect(document.getBlock(1)).toBe(block);
+    });
+
+    it('should add the block to the end of the document', () => {
```
neSpecc, line 107: name is not clear. What's the difference between this and `should add the block to the end of the document if index is not provided`?

src/entities/EditorDocument/EditorDocument.spec.ts
```diff
@@ -0,0 +1,228 @@
...
+      // Act
+      document.addBlock(block, document.length);
+
+      // Assert
+      const lastBlock = document.getBlock(document.length - 1);
+
+      expect(lastBlock).toBe(block);
+    });
+
+    it('should throw an error if index is greater then block nodes length', () => {
```
neSpecc, line 122:
```suggestion
    it('should throw an error if index is greater then document length', () => {
```

src/entities/EditorDocument/EditorDocument.spec.ts
```diff
@@ -0,0 +1,228 @@
...
+
+      // Act
+      const action = (): void => document.addBlock(block, -1);
+
+      // Assert
+      expect(action).toThrowError('Index out of bounds');
+    });
+  });
+
+  describe('removeBlock', () => {
```
neSpecc, line 149:
```suggestion
  describe('removeBlock()', () => {
```

src/entities/EditorDocument/EditorDocument.spec.ts
```diff
@@ -0,0 +1,228 @@
...
+    it('should remove the block from the beginning of the document', () => {
```
neSpecc, line 150:
```suggestion
    it('should remove the block from the beginning of the document if index=0 passed', () => {
```

src/entities/EditorDocument/EditorDocument.spec.ts
```diff
@@ -0,0 +1,228 @@
...
+      const documentLengthBeforeRemove = document.length;
+
+      // Act
+      document.removeBlock(document.length - 1);
+
+      // Assert
+      expect(document.length).toBe(documentLengthBeforeRemove - 1);
+    });
+
+    it('should throw an error if index is greater then block nodes length', () => {
```
neSpecc, line 183:
```suggestion
    it('should throw an error if index is greater then document length', () => {
```

src/entities/EditorDocument/EditorDocument.spec.ts
```diff
@@ -0,0 +1,228 @@
...
+    it('should throw an error if index is less then 0', () => {
+      // Act
+      const action = (): void => document.removeBlock(-1);
+
+      // Assert
+      expect(action).toThrowError('Index out of bounds');
+    });
+  });
+
+  describe('getBlock', () => {
```
neSpecc, line 200:
```suggestion
  describe('getBlock()', () => {
```

src/entities/EditorDocument/EditorDocument.spec.ts
```diff
@@ -0,0 +1,228 @@
+import { EditorDocument } from './index';
+import { BlockNode, BlockNodeName, createBlockNodeName } from '../BlockNode';
+
+const createBlock = ({ name, parent }: { name?: BlockNodeName, parent: EditorDocument }): BlockNode => {
+  return new BlockNode({
+    name: name || createBlockNodeName('block'),
```
neSpecc, line 6:
```ts
createBlockNodeName('block')
```

naming is not clear for me. What's happening here?

ilyamore88: I'm creating a block node name here. Because we use Nominal type for this field. I can't just use 'string' here.

src/entities/EditorDocument/EditorDocument.spec.ts
ilyamore88: Why do I need to duplicate the same code? The specific document isn't needed in most cases.

src/entities/EditorDocument/EditorDocument.spec.ts
neSpecc: so let's give more realistic name:

```ts
createBlockNodeName('header')
```

or at least 

```ts
createBlockNodeName('block-name-example')
```

src/entities/EditorDocument/EditorDocument.spec.ts
neSpecc: because test cases should not depend on each other. For example, the first test case mutates `document` variable. You can create `createDocumentMock(options)` helper to prevent code duplication

neSpecc (approved):

PR #6, opened 2023-05-25
feat(block-tune): implement update and serialised methods

ilyamore88:
I've implemented base methods for BlockTune and test cases for them.

### DIFF
--- a/src/entities/BlockTune/BlockTune.spec.ts
+++ b/src/entities/BlockTune/BlockTune.spec.ts
@@ -1,32 +1,13 @@
 import { BlockTune, BlockTuneName, createBlockTuneName } from './index';
-import { BlockNode, createBlockNodeName } from '../BlockNode';
-import { EditorDocument } from '../EditorDocument';
 
-const createBlockTune = ({ name, data, block }: {
+const createBlockTune = ({ name, data }: {
   name?: BlockTuneName,
-  data: Record<string, unknown>,
-  block?: BlockNode,
+  data?: Record<string, unknown>,
 }): BlockTune => {
-  const document = new EditorDocument({
-    children: [],
-    properties: {
-      readOnly: false,
-    },
-  });
-
-  const blockNode = block || new BlockNode({
-    name: createBlockNodeName('block'),
-    children: {},
-    parent: document,
-  });
-
-  document.addBlock(blockNode);
-
   return new BlockTune(
     {
       name: name || createBlockTuneName('blockTune'),
-      data,
-      block: blockNode,
+      data: data || {},
     }
   );
 };
--- a/src/entities/BlockTune/BlockTune.spec.ts
+++ b/src/entities/BlockTune/BlockTune.spec.ts
@@ -1,0 +1,94 @@
+import { BlockTune, BlockTuneName, createBlockTuneName } from './index';
+import { BlockNode, createBlockNodeName } from '../BlockNode';
+import { EditorDocument } from '../EditorDocument';
+
+const createBlockTune = ({ name, data, block }: {
+  name?: BlockTuneName,
+  data: Record<string, unknown>,
+  block?: BlockNode,
+}): BlockTune => {
+  const document = new EditorDocument({
+    children: [],
+    properties: {
+      readOnly: false,
+    },
+  });
+
+  const blockNode = block || new BlockNode({
+    name: createBlockNodeName('block'),
+    children: {},
+    parent: document,
+  });
+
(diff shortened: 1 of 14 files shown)


ilyamore88:
Removed `#block` field from BlockTune because it's unnecessary for now.

gohabereg (approved):

src/entities/BlockTune/BlockTune.spec.ts
```diff
@@ -0,0 +1,68 @@
+import { BlockTune, BlockTuneName, createBlockTuneName } from './index';
+
+const createBlockTune = ({ name, data }: {
```
neSpecc, line 3: its better to use named function instead of anonymous

```ts
function createBlockTune(){ ... }
```

src/entities/BlockTune/BlockTune.spec.ts
```diff
@@ -0,0 +1,68 @@
...
+  return new BlockTune(
+    {
+      name: name || createBlockTuneName('blockTune'),
+      data: data || {},
+    }
+  );
+};
+
+describe('BlockTune', () => {
+  describe('update method', () => {
```
neSpecc, line 16:
```suggestion
  describe('update()', () => {
```

src/entities/BlockTune/BlockTune.spec.ts
```diff
@@ -0,0 +1,68 @@
...
+    it('should add field to data object by key if it doesn\'t exist', () => {
+      // Arrange
+      const blockTune = createBlockTune({
+        data: {},
+      });
+
+      // Act
+      blockTune.update('test', 'test');
```
neSpecc, line 24: bad mocking values, please, use more real-life examples

src/entities/BlockTune/BlockTune.spec.ts
```diff
@@ -0,0 +1,68 @@
...
+      blockTune.update('test', 'newTest');
+
+      // Assert
+      expect(blockTune.serialized.data).toEqual({
+        test: 'newTest',
+      });
+    });
+  });
+
+  it('should return serialized version of the BlockTune', () => {
```
neSpecc, line 50: are you sure that `update()` should return that?

neSpecc: looks like it's a test for `.serizalized`

src/entities/BlockTune/index.ts
```diff
@@ -16,23 +15,36 @@ export class BlockTune {
    */
   #data: Record<string, unknown>;
```
neSpecc, line 16: maybe generic should be used here instead of `Record<string, unknown>`?

ilyamore88: Good question. I think we might use a generic type here. But for now, I'm not sure why we need to know the type of this data object in our document model. So we always can change it, but in a separate pr when we'll need it.

src/entities/BlockTune/BlockTune.spec.ts
```diff
@@ -0,0 +1,59 @@
...
+      };
+
+      // Act
+      const actual = createBlockTuneMock({
+        name: expected.name,
+        data: expected.data,
+      }).serialized;
+
+      // Assert
+      expect(actual).toEqual(expected);
```
neSpecc, line 56: its hard to read. It's bad practice to use "expected" as "arrange".

```suggestion
      // Arrange
      const tune = createBlockTuneMock({
        name: createBlockTuneName('styling'),
        data: {
          background: 'transparent',
        },
      });
      
      // Act
      const tuneSerialized = tune.serialized;
      
      // Assert
      expect(tuneSerialized).toEqual({
        name: "styling",
        data: {
          background: 'transparent',
        },
      });
```

src/entities/BlockTune/types/BlockTuneSerialized.ts
```diff
@@ -0,0 +1,16 @@
+import { BlockTuneName } from './BlockTuneName';
+
+/**
+ * BlockTuneSerialized represents a serialized version of a BlockTune.
+ */
+export interface BlockTuneSerialized {
+  /**
+   * The name of the tune.
+   */
+  name: BlockTuneName;
```
neSpecc, line 10: it should be a regular string after serialisation I guess?

ilyamore88: I agree with you. I'll fix it soon

neSpecc (approved):

PR #7, opened 2023-05-29
feat(value-node): update method, serialised and parent getters

ilyamore88:
Added ValueNode methods and getters with unit tests.


src/entities/ValueNode/ValueNode.spec.ts
```diff
@@ -0,0 +1,83 @@
+import { createValueNodeMock } from '../../utils/mocks/createValueNodeMock';
+import { BlockNode, createBlockNodeName } from '../BlockNode';
+import { createBlockNodeMock } from '../../utils/mocks/createBlockNodeMock';
+import { EditorDocument } from '../EditorDocument';
+
+describe('ValueNode', () => {
+  describe('.update()', () => {
+    it('should add new data field associated with this value node', () => {
```
neSpecc, line 8: Are you sure that we need to allow adding new properties to ValueNode? 

Seems like ValueNode should only allow updating of existed value which was set through the Constructor. And disallow adding new properties later

src/entities/BlockNode/types/BlockNodeChildren.ts
```diff
@@ -10,4 +10,4 @@ import { FormattingNode } from '../../FormattingNode';
  * The child is the ValueNode if the block node contains the data that will be fully replaced on a change event.
  * If the value is an array, it can contain formatting nodes or text nodes. Changes in these nodes works in a range.
  */
-export type BlockNodeChildren = Record<DataKey, ValueNode | (FormattingNode | TextNode)[]>;
+export type BlockNodeChildren = Record<DataKey, ValueNode[] | (FormattingNode | TextNode)[]>;
```
gohabereg, line 13: Why array?

src/entities/ValueNode/ValueNode.spec.ts
```diff
@@ -0,0 +1,66 @@
...
+import { createBlockNodeMock } from '../../utils/mocks/createBlockNodeMock';
+import { EditorDocument } from '../EditorDocument';
+import { createValueNodeName } from './types';
+
+describe('ValueNode', () => {
+  describe('.update()', () => {
+    it('should update existing data field associated with this value node', () => {
+      // Arrange
+      const valueNode = createValueNodeMock({
+        name: createValueNodeName('longitude'),
```
gohabereg, line 12: I don't think node itself needs to know it's DataKey

src/entities/ValueNode/index.ts
```diff
@@ -21,11 +26,44 @@ export class ValueNode {
...
+  public update(value: unknown): void {
+    this.#value = value;
+  }
+
+  /**
+   * Returns serialized data associated with this value node.
+   */
+  public get serialized(): ValueNodeSerialized {
+    return {
+      name: this.#name,
```
gohabereg, line 53: It should return just value

src/entities/ValueNode/ValueNode.spec.ts
gohabereg: Not sure about a parent as well though

gohabereg (changes requested):

src/utils/mocks/createValueNodeMock.ts
```diff
@@ -0,0 +1,13 @@
+import { ValueNode } from '../../entities/ValueNode';
+
+/**
+ * Creates a mock ValueNode instance.
+ *
+ * @param args - ValueNode constructor arguments.
+ * @param args.value - The value of this value node.
+ */
+export function createValueNodeMock({ value }: { value: unknown }): ValueNode {
```
gohabereg, line 9: You can use ConstructorParameters interface here

neSpecc (approved):
ok

gohabereg (approved):

PR #10, opened 2023-06-21
feat(block-node): add serialized method

ilyamore88:


ilyamore88:
There is one mutant that I can't kill without an implementation of `.serialized` methods for TextNode and FormattingNode.

src/entities/BlockNode/BlockNode.spec.ts
```diff
@@ -0,0 +1,65 @@
+import { createBlockNodeMock } from '../../utils/mocks/createBlockNodeMock';
+import { createBlockTuneMock } from '../../utils/mocks/createBlockTuneMock';
```
gohabereg, line 2: Those methods don't create mocks, they create real classes, so this test is actually an integration test.

For unit test, you should create mocked values and check that `serialized` getter is accessed for each mock

src/entities/BlockNode/BlockNode.spec.ts
```diff
@@ -0,0 +1,92 @@
...
+        data: {
+          [createDataKey('data-key-1a2b')]: textNodes,
+          ...valueNodes,
+        },
+        parent: {} as EditorDocument,
+      });
+
+      blockNode.serialized;
+
+      expect(textNodeSerializedSpy).toHaveBeenCalledTimes(countOfTextNodes);
```
gohabereg, line 88: Would be better to split this case into two: one for value nodes, one for text nodes

gohabereg (approved):

src/entities/BlockNode/BlockNode.spec.ts
```diff
@@ -0,0 +1,122 @@
...
+        parent: {} as EditorDocument,
+        tunes: blockTunesNames.reduce((acc, name) => ({
+          ...acc,
+          [name]: new BlockTune({} as BlockTuneConstructorParameters),
+        }), {}),
+      });
+
+      blockNode.serialized;
+
+      expect(spy).toHaveBeenCalledTimes(blockTunesNames.length);
```
neSpecc, line 72: can we check that `serialized` is called on exact passed tunes?

src/entities/BlockNode/index.ts
```diff
@@ -39,16 +40,55 @@ export class BlockNode {
...
+  /**
+   * Returns serialized object representing the BlockNode
+   */
+  public get serialized(): BlockNodeSerialized {
+    const serializedData = Object
+      .entries(this.#data)
+      .reduce(
+        (acc, [dataKey, value]) => {
+          if (value instanceof Array) {
+            acc[dataKey] = value.map((node) => node.serialized);
```
neSpecc, line 63: please add a comment here, maybe with the example of result key and value

neSpecc: also, please add a test case for it

ilyamore88: I've added a comment. There is a test case for TextNodes. I don't need to test the return value from node.serialized because it's a part of TextNode and FormattingNode.

neSpecc (approved):

PR #11, opened 2023-07-31
FormattingNode#unformat method & some tests refactoring

gohabereg:
- `unformat` method implementation
- mocks and tests refactoring

I'll do integration tests in a separate PR to keep everything smaller

### DIFF
--- a/src/entities/FormattingNode/index.ts
+++ b/src/entities/FormattingNode/index.ts
@@ -193,6 +193,8 @@ export class FormattingNode implements InlineNode {
   public format(tool: InlineToolName, start: number, end: number, data?: InlineToolData): InlineNode[] {
     /**
      * In case current tool is the same as new one, do nothing
+     *
+     * @todo Compare data as well
      */
     if (tool === this.#tool) {
       return [];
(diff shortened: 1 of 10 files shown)


src/entities/FormattingNode/index.ts
```diff
@@ -211,6 +213,51 @@ export class FormattingNode implements InlineNode {
...
+   * Removes formatting from the text for a specified inline tool in the specified range
+   *
+   * @param tool - name of inline tool to remove
+   * @param start - char start index of the range
+   * @param end - char end index of the range
+   * @todo Possibly pass data or some InlineTool identifier to relevant only required fragments
+   */
+  public unformat(tool: InlineToolName, start: number, end: number): InlineNode[] {
+    if (this.#tool === tool) {
+      const middleNode = this.split(start);
```
neSpecc, line 226: why split returns a single node? Maybe it should return an array of nodes? Because after splitting we have 2 nodes

gohabereg: That's because split method creates only one new node, you already have a link to the original one:

```
const newNode = node.split(...);

const leftPart = node;
const rightPart = newNode;
```

So no need to make output more complex

neSpecc (approved):

PR #12, opened 2023-08-24
RootInlineNode

gohabereg:
PR adds:
- RootInlineNode as single point access to inline fragment tree
- Updates directory structure
- Add tree normalization on input and output
- Inline fragments tree integration tests (done, but all tests require refactoring)


src/entities/inline-fragments/specs/InlineTree.integration.spec.ts
```diff
@@ -0,0 +1,373 @@
+import { RootInlineNode } from '../RootInlineNode';
+import { TextNode } from '../TextNode';
+import { createInlineToolData, createInlineToolName } from '../FormattingNode';
+
+describe('Inline fragments tree integration', () => {
```
neSpecc, line 5:
```suggestion
describe('Inline fragments tree integration', () => {
```

src/entities/inline-fragments/specs/InlineTree.integration.spec.ts
```diff
@@ -0,0 +1,373 @@
...
+      const index = 10;
+      const length = 5;
+
+      tree.removeText(index, index + length);
+
+      expect(tree.getText())
+        .toBe(initialText.slice(0, index) + initialText.slice(index + length));
+    });
+
+    it('should remove text from the end of the tree', () => {
```
neSpecc, line 106:
```suggestion
    it('should remove text from passed index to the end of the tree', () => {
```

src/entities/inline-fragments/specs/InlineTree.integration.spec.ts
```diff
@@ -0,0 +1,373 @@
...
+    });
+
+    it('should throw an error if index is out of range', () => {
+      const initialText = 'Editor.js is a block-styled editor.';
+      const child = new TextNode({ value: initialText });
+      const tree = new RootInlineNode({ children: [ child ] });
+      const index = 100;
+
+      expect(() => tree.removeText(index))
+        .toThrowError();
```
neSpecc, line 125: could we also check the error text or type here?

src/entities/inline-fragments/specs/InlineTree.integration.spec.ts
```diff
@@ -0,0 +1,373 @@
...
+          range: [
+            index + offset,
+            index + length + offset,
+          ],
+        },
+      ]);
+    });
+  });
+
+  it('should pass "ultimate" test', () => {
```
neSpecc, line 310: please add a few words of what is happening and asserting here

src/entities/inline-fragments/TextNode/index.ts
```diff
@@ -14,7 +15,7 @@ export class TextNode implements InlineNode {
   /**
    * Private field representing the text content of the node
```
ilyamore88, line 16: Fix doc here

src/entities/inline-fragments/FormattingNode/index.ts
```diff
@@ -0,0 +1,223 @@
...
+      currentFragment.data = this.data;
+    }
+
+    fragments.unshift(currentFragment);
+
+    return fragments;
+  }
+
+  /**
+   * Splits current node by the specified index
```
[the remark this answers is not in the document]
gohabereg: split is what this method do — splits node into two ones. And it used in the cases where we need to _split_ code. I understand your concerns, but split for array is not split for a tree node

src/entities/inline-fragments/mixins/ParentNode/index.ts
```diff
@@ -142,5 +154,35 @@ export function ParentNode<C extends { new(...args: any[]): InlineNode }>(constr
...
+     */
+    public normalize(): void {
+      const children = Array.from(this.children);
+
+      if (!children.length) {
+        return;
+      }
+
+      children.reduce((prev, child) => {
+        if (prev && prev.isEqual(child)) {
```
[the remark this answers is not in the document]
gohabereg: No, if data is not equal we don't need to merge nodes — as we won't be able to persist data for both of them

src/entities/inline-fragments/specs/InlineTree.integration.spec.ts
gohabereg: It seems you forgot your suggestion here

src/entities/inline-fragments/specs/InlineTree.integration.spec.ts
gohabereg: It's not a good idea to check for error text as we might change it later. We can use custom error classes, but that's a whole big task

src/entities/inline-fragments/FormattingNode/index.ts
```diff
@@ -0,0 +1,223 @@
...
+   *
+   * @param node - node to merge with
+   */
+  public mergeWith(node: InlineNode): void {
+    if (!this.isEqual(node)) {
+      throw new Error('Can not merge unequal nodes');
+    }
+
+    /**
+     * @todo merge data
```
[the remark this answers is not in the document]
gohabereg: That should be decided by the tool. There might be cases where data is not equal but still mergable

src/entities/inline-fragments/ParentInlineNode/index.ts
```diff
@@ -1,50 +1,29 @@
...
  */
 @ParentNode
-@ChildNode
-export class FormattingNode implements InlineNode {
-  /**
-   * Private field representing the name of the formatting tool applied to the content
-   */
-  #tool: InlineToolName;
-
+export class ParentInlineNode implements InlineNode {
```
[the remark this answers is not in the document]
gohabereg: Abstract here breaks the types. Anyway, there is no abstract methods in the class, an you can actually create an instance of ParentInlineNode — it doesn't break any logic

src/entities/BlockNode/index.ts
```diff
@@ -38,7 +38,7 @@ export class BlockNode {
   /**
    * Constructor for BlockNode class.
    *
-   * @param args - TextNode constructor arguments.
+   * @param args - TextInlineNode constructor arguments.
```
ilyamore88, line 41:
```suggestion
   * @param args - BlockNode constructor arguments.
```

src/entities/inline-fragments/FormattingInlineNode/types/InlineToolName.ts
ilyamore88, line 1: Update docs in this file

ilyamore88 (approved):

src/entities/inline-fragments/specs/InlineTree.integration.spec.ts
neSpecc: If we change the error text, we can also change it in tests. For now any error thrown during execution of `removeText()` will lead false-positive test assertion

neSpecc (approved):

PR #15, opened 2023-08-27
feat(editor-document): add .properties getter, add .updateProperty method

ilyamore88:


src/entities/EditorDocument/index.ts
```diff
@@ -80,6 +80,23 @@ export class EditorDocument {
...
+    return this.#properties;
+  }
+
+  /**
+   * Updates a property of the EditorDocument.
+   *
+   * @param name - The name of the property to update
+   * @param value - The value to update the property with
+   */
+  public updateProperty(name: PropName, value: unknown): void {
```
neSpecc, line 96: maybe setProperty?

src/entities/EditorDocument/index.ts
```diff
@@ -80,6 +80,23 @@ export class EditorDocument {
...
+    this.#properties[name] = value;
```
neSpecc, line 97: could you write a test case of passing unexisting property name?

ilyamore88: Done

gohabereg (approved):

src/entities/EditorDocument/EditorDocument.spec.ts
```diff
@@ -246,4 +246,80 @@ describe('EditorDocument', () => {
...
+
+      expect(document.properties[propertyName]).toBe(expectedValue);
+    });
+
+    it('should add the property if it does not exist', () => {
+      const propertyName = 'readOnly';
+      const expectedValue = true;
+      const document = new EditorDocument({
+        children: [],
+        properties: {},
```
neSpecc, line 317: I think, `properties` should be optional in construction params. And set as an empty object in a constructor.

ilyamore88: Made it optional. But for these test cases, I want to have an empty object in the constructor for more readability.

neSpecc (approved):

PR #39, opened 2023-10-05
Implement DocumentModel initialization

gohabereg:
### Changes

1. Update entities to accept serialized data into constructor — it enables model initialization using saved data
2. Update BlockNodeData structure to accept any object/array structures as Tool data
3. Update serialized data schema to mark ValueNode/TextNode data with special property `$t`
4. Add private `initialize` methods to EditorDocument and BlockNode classes to create entities from passed serialized data
5. Update ValueNode/TextNode methods call from BlockNode to support any tool data structure

Two issues raised to cover some of added code with integration tests: #40 #41

### DIFF
--- a/src/entities/BlockNode/BlockNode.spec.ts
+++ b/src/entities/BlockNode/BlockNode.spec.ts
@@ -6,7 +6,7 @@ import { ValueNode } from '../ValueNode';
 import type { EditorDocument } from '../EditorDocument';
 import type { ValueNodeConstructorParameters } from '../ValueNode';
 import { InlineToolData, InlineToolName, TextNode } from '../inline-fragments';
-import { BlockChildType } from './types';
+import { BlockChildType, BlockNodeDataSerialized } from './types';
 import { NODE_TYPE_HIDDEN_PROP } from './consts';
 
 jest.mock('../BlockTune');
@@ -14,6 +14,13 @@ jest.mock('../BlockTune');
 jest.mock('../inline-fragments/TextNode');
 
 jest.mock('../ValueNode');
+
+const createBlockNodeWithData = (data: BlockNodeDataSerialized): BlockNode => {
+  return new BlockNode({
+    name: createBlockToolName('header'),
+    data,
+  });
+};
 
 describe('BlockNode', () => {
   describe('constructor', () => {
@@ -513,46 +520,18 @@ describe('BlockNode', () => {
   });
 
   describe('.insertText()', () => {
-    let node: BlockNode;
     const dataKey = createDataKey('text');
     const text = 'Some text';
 
-    beforeEach(() => {
-      node = new BlockNode({
-        name: createBlockToolName('header'),
-        data: {
-          [dataKey]: {
-            [NODE_TYPE_HIDDEN_PROP]: BlockChildType.Text,
-            value: '',
-            fragments: [],
-          },
-          object: {
-            [dataKey]: {
-              [NODE_TYPE_HIDDEN_PROP]: BlockChildType.Text,
-              value: '',
-              fragments: [],
-            },
-            array: [
-              {
-                [NODE_TYPE_HIDDEN_PROP]: BlockChildType.Text,
-                value: '',
-                fragments: [],
-              },
-            ],
-          },
-          array: [
-            {
-              [NODE_TYPE_HIDDEN_PROP]: BlockChildType.Text,
-              value: '',
-              fragments: [],
-            },
-          ],
-        },
-      });
-    });
-
     it('should call .insertText() method of the TextNode', () => {
       const spy = jest.spyOn(TextNode.prototype, 'insertText');
+      const node = createBlockNodeWithData({
+        [dataKey]: {
+          [NODE_TYPE_HIDDEN_PROP]: BlockChildType.Text,
+          value: '',
+          fragments: [],
+        },
+      });
 
       node.insertText(dataKey, text);
 
@@ -562,6 +541,15 @@ describe('BlockNode', () => {
 
     it('should call .insertText() method of the TextNode in an object', () => {
       const spy = jest.spyOn(TextNode.prototype, 'insertText');
+      const node = createBlockNodeWithData({
+        object: {
+          [dataKey]: {
+            [NODE_TYPE_HIDDEN_PROP]: BlockChildType.Text,
+            value: '',
+            fragments: [],
+          },
+        },
+      });
 
       node.insertText(createDataKey(`object.${dataKey}`), text);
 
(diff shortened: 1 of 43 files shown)


src/entities/BlockNode/BlockNode.spec.ts
```diff
@@ -137,9 +130,228 @@ describe('BlockNode', () => {
...
+        parent: {} as EditorDocument,
+      });
+
+      blockNode.serialized;
+
+      expect(spy)
+        .toHaveBeenCalledTimes(1);
+    });
+
+    it('should call .serialized getter of ValueNodes in an nested object', () => {
```
ilyamore88, line 176:
```suggestion
    it('should call .serialized getter of ValueNodes in a nested object', () => {
```

src/entities/BlockNode/BlockNode.spec.ts
```diff
@@ -137,9 +130,228 @@ describe('BlockNode', () => {
...
+      const blockNode = new BlockNode({
+        name: createBlockToolName('paragraph'),
+        data: {
+          array: [ {
+            object: {
+              text: {
+                $t: BlockChildType.Text,
+                value: '',
+                fragments: [],
+              },
```
ilyamore88, line 287: I thought we decided to detect text nodes by value and fragment fields. Is $t field required here?

neSpecc:
Could you please write a short changelog in the PR description. It helps with review

src/entities/ValueNode/types/ValueSerialized.ts
```diff
@@ -0,0 +1,6 @@
+import { BlockChildType } from '../../BlockNode/types';
+
+/**
+ * Type representing serialized ValueNode
```
neSpecc, line 4: It's better to describe what actually means to be a value node serialized. Eg. "Serialized ValueNode presented via JSON with the $t property which is used to determine that this structure is a ValueNode"

src/entities/ValueNode/index.ts
```diff
@@ -33,8 +34,14 @@ export class ValueNode<ValueType = unknown> {
   /**
    * Returns serialized data associated with this value node.
    */
-  public get serialized(): ValueType {
-    return this.#value;
+  public get serialized(): ValueSerialized<ValueType> {
+    let value = this.#value;
+
+    if (typeof value === 'object' && this.#value !== null) {
+      value = Object.assign({ $t: BlockChildType.Value }, value);
```
neSpecc, line 41: maybe it would be useful to move "$t" to some constant with clear name like "NODE_TYPE_HIDDEN_PROP"?

src/entities/ValueNode/ValueNode.spec.ts
```diff
@@ -31,5 +32,16 @@ describe('ValueNode', () => {
       // Assert
       expect(serializedLongitude).toStrictEqual(longitude);
     });
+
+    it('should mark serialized value as value node if object returned', () => {
```
neSpecc, line 36:
```suggestion
    it('should mark serialized value as value node by using custom hidden property $t if object returned', () => {
```

src/entities/EditorDocument/types/EditorDocumentSerialized.ts
```diff
@@ -0,0 +1,7 @@
+import { BlockNodeSerialized } from '../../BlockNode/types';
+import { Properties } from './Properties';
+
+export interface EditorDocumentSerialized {
+    blocks: BlockNodeSerialized[];
+    properties: Properties;
```
neSpecc, line 6: pls add a description, what is stored here

src/entities/EditorDocument/index.ts
```diff
@@ -209,6 +211,27 @@ export class EditorDocument {
     this.#children[blockIndex].unformat(key, tool, start, end);
   }
 
+  /**
+   *
```
neSpecc, line 215: Describe how we decided to serialise the Document

src/entities/EditorDocument/EditorDocument.spec.ts
```diff
@@ -1,6 +1,5 @@
 import { EditorDocument } from './index';
 import { BlockNode, BlockToolName, DataKey } from '../BlockNode';
```
neSpecc, line 2: Looks like this test should not depend on BlockNode types and operate only with serialised values

src/entities/BlockNode/types/DataKey.ts
```diff
@@ -3,7 +3,7 @@ import { create, Nominal } from '../../../utils/Nominal';
 /**
  * Base type of the data key field
  */
-type DataKeyBase = string;
+type DataKeyBase = string | number;
```
neSpecc, line 6: pls describe in jsdoc when there is a string and when a number

src/entities/BlockNode/types/BlockNodeSerialized.ts
```diff
@@ -1,4 +1,24 @@
...
+
+/**
+ * Union type of serialized BlockNode child nodes
+ */
+export type BlockChildNodeSerialized = ValueSerialized | TextNodeSerialized;
+
+/**
+ * Reccurrent type representing serialized BlockNode data
+ */
+export type BlockNodeDataSerializedValue = BlockChildNodeSerialized | BlockChildNodeSerialized[] | BlockNodeDataSerialized | BlockNodeDataSerialized[];
```
neSpecc, line 13: maybe `BlockNodeDataValueSerialized`?

src/entities/BlockNode/types/BlockChildType.ts
```diff
@@ -0,0 +1,4 @@
+export enum BlockChildType {
+  Value = 'v',
+  Text = 't',
```
neSpecc, line 3: do we need to add type property to all types? Maybe only to ValueNodes would be enough?

src/entities/BlockNode/index.ts
```diff
@@ -174,6 +207,42 @@ export class BlockNode {
...
   }
 
+  /**
+   * Initializes BlockNode with passed block data
+   *
+   * @param data - block data
+   */
+  #initialize(data: BlockNodeDataSerialized): void {
+    const map = (value: BlockNodeDataSerializedValue): BlockNodeData | BlockNodeDataValue => {
+      if (Array.isArray(value)) {
```
neSpecc, line 217: pld add comments with examples to main cases

src/entities/BlockNode/index.ts
```diff
@@ -64,10 +82,25 @@ export class BlockNode {
...
+      }
+
+      if (data instanceof ValueNode || data instanceof TextNode) {
+        return data.serialized;
+      }
+
+      return Object.fromEntries(
+        Object.entries(data)
+          .map(([key, value]) => ([key, map(value)]))
+      );
```
neSpecc, line 97: looks hard to read, maybe it would be better to move it to utility with a nice name?

src/entities/BlockNode/BlockNode.spec.ts
gohabereg: I think it would be more consistent to add $t property for text nodes as well. So in case we'd like to change the data structure of a text node, we could still detect it using $t prop

src/entities/EditorDocument/EditorDocument.spec.ts
gohabereg: These dependencies are for types and assertions

src/entities/BlockNode/types/BlockChildType.ts
gohabereg: See https://github.com/editor-js/document-model/pull/39#discussion_r1370877258

neSpecc (approved):

ilyamore88 (approved):

PR #42, opened 2023-11-09
deps(yarn): update yarn to latest version

ilyamore88:
# Changelog
- Update Yarn to v4.0.1
- Add .nvmrc file with Node LTS v20.9.0
- Update actions/checkout to v4
- Fix warning about using `::set-output`


neSpecc (approved):
seems legit

gohabereg (approved):

Issue #22, opened 2023-08-29
Create an EventBus class


gohabereg:
1. Create an EventBus class
2. Create class or method which would create an event
3. Find a way to pass EventBus to Model nodes (DI?)

gohabereg:
updated
```{
  index: `${DocumentId}:${BlockId}:data@${DataKey}|tune@${TuneName}:${StartIndex}`
  event: ‘added’ | ‘removed’ | ‘modified’,  
  data: 
    text (added, removed) -> new text value
    text (modified) -> end index, new formatting name + value, isInverted
    value (modified) -> new value
    tune (modified) -> new tune value
    block -> BlockSerialized
    document (modified) -> prop name , prop value
}
```

gohabereg:
Closed by #43

PR #43, opened 2023-11-09
feat(event-bus): implement simple eventbus class

ilyamore88:
Task #22 

# Changelog
- Add EventBus class
- EventBus extends EventTarget methods
- Add base events for blocks

# ToDo
- Add types for event objects
- Connect the Document model to EventBus


gohabereg (approved):

src/utils/EventBus/EventBus.ts
```diff
@@ -0,0 +1,64 @@
+import EventEmitter from 'eventemitter3';
```
neSpecc, line 1: why not to use native `EventTarget`?

neSpecc: 70+ kb is too much for such a simple thing, I guess.

ilyamore88: Replaced with EventTarget

src/utils/EventBus/types/EventType.ts
```diff
@@ -0,0 +1,9 @@
+/**
+ * Event types enum that can be emitted by the document model
+ */
+export enum EventType {
+  /**
+   * The document model has been changed
+   */
+  CHANGED = 'changed',
```
gohabereg, line 8: Please use PascalCase

gohabereg (approved):

neSpecc (approved):

PR #52, opened 2023-11-15
feat(playground): document tree rendering

neSpecc:
Now we have a base Editor Document tree rendering

image


packages/model/src/entities/BlockNode/index.ts
```diff
@@ -78,6 +78,21 @@ export class BlockNode {
...
+   * Allows accessing Block name
+   */
+  public get name(): string {
+    return this.#name;
+  }
+
+  /**
+   * Allows accessing Block data
+   */
+  public get data(): BlockNodeData {
```
gohabereg, line 91: We should make the return type readonly (would be nice to check if it works for nested structures as well)

packages/model/src/entities/ValueNode/index.ts
```diff
@@ -44,6 +44,13 @@ export class ValueNode<ValueType = unknown> {
 
     return value as ValueSerialized<ValueType>;
   }
+
+  /**
+   * Returns the data associated with this value node.
+   */
+  public get value(): ValueType {
```
gohabereg, line 51: Also should be readonly as it could be array/object

packages/playground/src/components/Value.vue
```diff
@@ -0,0 +1,59 @@
...
+defineProps<{
+  value: object | null | number | string | boolean | Array<unknown>;
+}>();
+
+/**
+ * Returns whether the value is an object
+ *
+ * @param value - The value to check
+ */
+function isObject(value: unknown): value is object {
```
gohabereg, line 13: Could be extracted to a separate function

packages/model/src/utils/DeepReadonly.ts
```diff
@@ -0,0 +1,15 @@
+/**
+ * Make all properties in T readonly
+ */
+export type DeepReadonly<T> =
+    T extends (infer R)[] ? DeepReadonlyArray<R> :
+    // eslint-disable-next-line @typescript-eslint/ban-types
```
gohabereg, line 6: You can remove this if you use `(...args: unknown[]) => unknown` instead of Function

packages/playground/src/components/Node.vue
```diff
@@ -0,0 +1,112 @@
...
+ */
+function getClassProperties<T extends object>(object: T): PropsList<T> {
+  const descriptors = Object.getOwnPropertyDescriptors(object);
+
+  return Object.entries(descriptors).filter(([name, descriptor]) => {
+    return descriptor.get !== undefined;
+  })
+    .map(([ name ]) => {
+      return name as keyof typeof props.node;
+    });
```
gohabereg, line 38: Looks like some formatting could be applied here

gohabereg (approved):

ilyamore88 (approved):

PR #54, opened 2023-11-15
feat(block-node): add getFragments method

ilyamore88:


packages/model/src/entities/BlockNode/BlockNode.integration.spec.ts
```diff
@@ -100,4 +102,114 @@ describe('BlockNode integration tests', () => {
...
+        name: 'blockNode',
+        data: {
+          [dataKey]: {
+            [NODE_TYPE_HIDDEN_PROP]: BlockChildType.Text,
+            value: 'value',
+          },
+        },
+      });
+
+      const fragments= node.getFragments(
```
gohabereg, line 122: We need prettier

packages/model/src/entities/inline-fragments/FormattingInlineNode/index.ts
```diff
@@ -65,8 +65,7 @@ export class FormattingInlineNode extends ParentInlineNode implements InlineNode
 
   /**
    * Returns inline fragments for node from the specified character range
-   *
-   * If start and/or end is specified, method will return partial fragments for the specified range
+   * Always return fragments with whole range even if it's not fully covered by the passed range
```
gohabereg, line 68:
```suggestion
   * Always return full fragments even if one is not fully covered by the passed range
```

packages/model/src/entities/BlockNode/index.ts
```diff
@@ -205,6 +205,30 @@ export class BlockNode {
...
+  public getFragments(dataKey: DataKey, start: number, end: number, tool?: InlineToolName): InlineFragment[] {
+    this.#validateKey(dataKey, TextNode);
+
+    const node = get<TextNode>(this.#data, dataKey as string)!;
+
+    let fragments = node.getFragments(start, end);
+
+    if (tool) {
+      fragments = fragments.filter(fragment => fragment.tool === tool);
+    }
```
gohabereg, line 227: Would be better to filter inside TextNode

gohabereg (commented):
Please propagate the method through the whole structure

neSpecc (approved):

gohabereg (approved):

PR #58, opened 2023-11-18
feat(adapters): caret adapter basic implementation

neSpecc:
Basic implementation of the Caret DOM Adapter

- Binds "selectionchange" event on document using "useSelectionChange" util providing "on" and "off" methods.
- "useSelectionChange" creates a singleton and delegates listeners
- Caret Adapter provides `attachInput` and `detachInput` methods
- Caret Adapter extends EventTarget and emits the `change` event — used for playground purposes. Could be removed later
- Playground contains `Input` components with Caret Adapter attached. Caret index is displayed at the right side of the input

image

### DIFF
--- a/packages/dom-adapters/src/caret/CaretAdapter.ts
+++ b/packages/dom-adapters/src/caret/CaretAdapter.ts
@@ -18,7 +18,7 @@ export class CaretAdapter extends EventTarget {
   /**
    * Index stores start and end offset of a caret depending on a root of input
    */
-  #index: TextRange = [0, 0];
+  #index: TextRange | null = null;
 
   /**
    * Input element
--- a/packages/dom-adapters/src/caret/CaretAdapter.ts
+++ b/packages/dom-adapters/src/caret/CaretAdapter.ts
@@ -1,5 +1,5 @@
 import type { EditorJSModel, TextRange } from '@editorjs/model';
-import { useSelectionChange, type Subscriber  } from './utils/useSelectionChange.js';
+import { useSelectionChange, type Subscriber, type InputWithCaret  } from './utils/useSelectionChange.js';
 import { getAbsoluteRangeOffset } from './utils/absoluteOffset.js';
 
 /**
@@ -23,38 +23,37 @@ export class CaretAdapter extends EventTarget {
   /**
    * Input element
    */
-  #input: null | HTMLElement = null;
+  #input: null | InputWithCaret = null;
+
+  /**
+   * EditorJSModel instance
+   */
+  #model: EditorJSModel;
+
+  /**
+   * Index of a block that contains input
+   */
+  #blockIndex: number;
 
   /**
    * Method for subscribing on document selection change
    */
-  #onSelectionChange: (callback: Subscriber) => void;
+  #onSelectionChange: (input: InputWithCaret, callback: Subscriber['callback'], context: Subscriber['context']) => void;
 
   /**
    * Method for unsubscribing from document selection change
    */
-  #offSelectionChange: (callback: Subscriber) => void;
-
-  /**
-   * Callback that will be called on document selection change
-   * Stored as a class property to be able to unsubscribe from document selection change
-   *
-   * @param selection - changed document selection
-   */
-  #onDocumentSelectionChange = (selection: Selection | null): void => {
-    if (!this.#isSelectionRelatedToInput(selection)) {
-      return;
-    }
-
-    this.#updateIndex(selection);
-  };
+  #offSelectionChange: (input: InputWithCaret) => void;
 
   /**
    * @param model - EditorJSModel instance
    * @param blockIndex - index of a block that contains input
    */
-  constructor(private readonly model: EditorJSModel, private readonly blockIndex: number) {
+  constructor(model: EditorJSModel, blockIndex: number) {
     super();
+
+    this.#model = model;
+    this.#blockIndex = blockIndex;
 
     const { on, off } = useSelectionChange();
 
@@ -72,30 +71,29 @@ export class CaretAdapter extends EventTarget {
   public attachInput(input: HTMLElement, dataKey: string): void {
     this.#input = input;
 
-    this.#onSelectionChange(this.#onDocumentSelectionChange);
+    this.#onSelectionChange(this.#input, this.#onDocumentSelectionChange, this);
   }
 
   /**
    * Unsubscribes from input caret change
    */
   public detachInput(): void {
-    this.#offSelectionChange(this.#onDocumentSelectionChange);
+    if (!this.#input) {
+      return;
+    }
+
+    this.#offSelectionChange(this.#input);
+    this.#input = null;
   }
 
   /**
-   * Checks if selection is related to input
+   * Callback that will be called on document selection change
    *
    * @param selection - changed document selection
    */
-  #isSelectionRelatedToInput(selection: Selection | null): boolean {
-    if (!selection) {
-      return false;
-    }
-
-    const range = selection.getRangeAt(0);
-
-    return this.#input?.contains(range.startContainer) ?? false;
-  }
+  #onDocumentSelectionChange(selection: Selection | null): void {
+    this.#updateIndex(selection);
+  };
 
   /**
    * Returns absolute caret index related to input
--- a/packages/dom-adapters/src/caret/CaretAdapter.ts
+++ b/packages/dom-adapters/src/caret/CaretAdapter.ts
@@ -71,7 +71,7 @@ export class CaretAdapter extends EventTarget {
   public attachInput(input: HTMLElement, dataKey: string): void {
     this.#input = input;
 
-    this.#onSelectionChange(this.#input, this.#onDocumentSelectionChange, this);
+    this.#onSelectionChange(this.#input, this.#onInputSelectionChange, this);
   }
 
   /**
@@ -91,7 +91,7 @@ export class CaretAdapter extends EventTarget {
    *
    * @param selection - changed document selection
    */
-  #onDocumentSelectionChange(selection: Selection | null): void {
+  #onInputSelectionChange(selection: Selection | null): void {
     this.#updateIndex(selection);
   };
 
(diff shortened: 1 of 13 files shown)


packages/dom-adapters/src/caret/CaretAdapter.ts
```diff
@@ -0,0 +1,140 @@
+import type { EditorJSModel } from '@editorjs/model';
+import { useSelectionChange, type Subscriber  } from './utils/useSelectionChange.js';
+import { getAbsoluteRangeOffset } from './utils/absoluteOffset.js';
+
+/**
+ * Caret index is a tuple of start and end offset of a caret
+ */
+export type CaretIndex = [number, number];
```
gohabereg, line 8: We might re-use type from model

packages/dom-adapters/src/caret/CaretAdapter.ts
```diff
@@ -0,0 +1,140 @@
...
+ *
+ * @todo add support for native inputs
+ * @todo debug problem when document "selectionchange" is not fired on Enter press
+ * @todo debug problem when offset at the end of line and at the beginning of the next line is the same
+ */
+export class CaretAdapter extends EventTarget {
+  /**
+   * Index stores start and end offset of a caret depending on a root of input
+   */
+  #index: CaretIndex = [0, 0];
```
gohabereg, line 26: If caret is not set, what index it would be?

packages/dom-adapters/src/caret/CaretAdapter.ts
```diff
@@ -0,0 +1,140 @@
...
+   */
+  #offSelectionChange: (callback: Subscriber) => void;
+
+  /**
+   * Callback that will be called on document selection change
+   * Stored as a class property to be able to unsubscribe from document selection change
+   *
+   * @param selection - changed document selection
+   */
+  #onDocumentSelectionChange = (selection: Selection | null): void => {
```
gohabereg, line 49: We could try to optimise memory usage with a method (not a property). And pass the context from outside

packages/dom-adapters/src/caret/CaretAdapter.ts
```diff
@@ -0,0 +1,140 @@
...
+    }
+
+    this.#updateIndex(selection);
+  };
+
+  /**
+   * @param model - EditorJSModel instance
+   * @param blockIndex - index of a block that contains input
+   */
+  constructor(private readonly model: EditorJSModel, private readonly blockIndex: number) {
```
gohabereg, line 61: Would be better to use # modifier. It would add more lines, but properties would be "really" private

packages/dom-adapters/src/caret/CaretAdapter.ts
```diff
@@ -0,0 +1,140 @@
...
+  }
+
+  /**
+   * - Subscribes on input caret change
+   * - Composes caret index by selection position related to start of input
+   *
+   * @param input - input to watch caret change
+   * @param dataKey - key of data property in block's data that contains input's value
+   */
+  public attachInput(input: HTMLElement, dataKey: string): void {
```
gohabereg, line 77: You need an index of the block as well. Also dataKey is not used

packages/dom-adapters/src/caret/CaretAdapter.ts
```diff
@@ -0,0 +1,140 @@
...
+    this.#input = input;
+
+    this.#onSelectionChange(this.#onDocumentSelectionChange);
+  }
+
+  /**
+   * Unsubscribes from input caret change
+   */
+  public detachInput(): void {
+    this.#offSelectionChange(this.#onDocumentSelectionChange);
```
gohabereg, line 87: You could set null to #input

packages/dom-adapters/src/caret/CaretAdapter.ts
```diff
@@ -0,0 +1,140 @@
...
+   * @param selection - changed document selection
+   */
+  #isSelectionRelatedToInput(selection: Selection | null): boolean {
+    if (!selection) {
+      return false;
+    }
+
+    const range = selection.getRangeAt(0);
+
+    return this.#input?.contains(range.startContainer) ?? false;
```
gohabereg, line 102: Ok for now, but we should think of cross-block selection case in the future

packages/dom-adapters/src/caret/utils/absoluteOffset.ts
```diff
@@ -0,0 +1,57 @@
...
+ *
+ * <div>
+ *   <div>
+ *     <br>
+ *   </div>
+ * </div>
+ *
+ * @param node - node to check
+ */
+function isLineBreak(node: Node): boolean {
```
gohabereg, line 12: Could you move it below the exported members

packages/dom-adapters/src/caret/utils/useSelectionChange.ts
```diff
@@ -0,0 +1,51 @@
...
+  (selection: Selection | null): void;
+}
+
+/**
+ * Utility composable that watches for document "selection change" event and delegates the provided callbacks to subscribers.
+ */
+export const useSelectionChange = createSingleton(() => {
+  const subscribers = new Set<Subscriber>();
+
+  document.addEventListener('selectionchange', () => {
```
gohabereg, line 21: Do we need an API to remove the listener?

packages/dom-adapters/src/caret/CaretAdapter.ts
neSpecc: block index passed to the constructor. It is not used right now as well as `model` and `dataKey`. I'll implement a working with model later

packages/dom-adapters/src/caret/utils/absoluteOffset.ts
neSpecc: There will be ES Lint error: using of method before it declared

packages/dom-adapters/src/caret/CaretAdapter.ts
```diff
@@ -0,0 +1,133 @@
...
+   * @param selection - changed document selection
+   */
+  #onInputSelectionChange(selection: Selection | null): void {
+    this.#updateIndex(selection);
+  };
+
+  /**
+   * Returns absolute caret index related to input
+   */
+  public get index(): TextRange {
```
gohabereg, line 101: Should be above private methods

packages/dom-adapters/src/caret/utils/useSelectionChange.ts
```diff
@@ -0,0 +1,143 @@
...
+   * Callback that will be called on "selection change" event.
+   *
+   * @param selection - current document selection
+   */
+  callback: (selection: Selection | null) => void;
+
+  /**
+   * Used to save context of the callback.
+   */
+  context: CaretAdapter;
```
gohabereg, line 18: It might be not only CaretAdapter in the future I suppose. Could be just unknown

packages/dom-adapters/src/caret/utils/useSelectionChange.ts
```diff
@@ -0,0 +1,143 @@
...
+   * @param input - input to check
+   */
+  function isSelectionRelatedToInput(selection: Selection | null, input: InputWithCaret): boolean {
+    if (!selection) {
+      return false;
+    }
+
+    const range = selection.getRangeAt(0);
+
+    return input.contains(range.startContainer) ?? false;
```
gohabereg, line 58: Could you just add todo to think of cross-block selection later. To know where to look once we'll start working on it

packages/dom-adapters/src/caret/utils/useSelectionChange.ts
```diff
@@ -0,0 +1,143 @@
...
+ */
+export type InputWithCaret = HTMLElement;
+
+
+/**
+ * Utility composable that watches for document "selection change" event and delegates the provided callbacks to subscribers.
+ */
+export const useSelectionChange = createSingleton(() => {
+  /**
+   * User to iterate over all inputs and check if selection is related to them.
```
gohabereg, line 36:
```suggestion
   * Used to iterate over all inputs and check if selection is related to them.
```

packages/dom-adapters/src/caret/utils/useSelectionChange.ts
```diff
@@ -0,0 +1,143 @@
...
+   */
+  const inputsWatched: InputWithCaret[] = [];
+
+  /**
+   * WeakMap that stores subscribers for each input.
+   */
+  const subscribers = new WeakMap<InputWithCaret, Subscriber>();
```
gohabereg, line 43: Could there be more than one subscriber for an input?

packages/dom-adapters/src/caret/utils/useSelectionChange.ts
```diff
@@ -0,0 +1,143 @@
...
+      }
+    });
+  }
+
+  init();
+
+  return {
+    on,
+    off,
+    destroy,
```
gohabereg, line 141: Once destroyed, there'll be no way to re-init as `createSingletone` will always return the same instance

packages/dom-adapters/src/caret/utils/useSelectionChange.ts
gohabereg: You can actually use `range.intersectsNode(input)`

packages/dom-adapters/src/caret/utils/useSelectionChange.ts
neSpecc: no

packages/dom-adapters/src/caret/utils/useSelectionChange.ts
neSpecc: yes

neSpecc: I've exported `init` method as well. The destroy flow is not designed yet, so it will be implemented once we can understand it

gohabereg (approved):

ilyamore88 (approved):

PR #65, opened 2023-12-23
feat(block-tool-adapter): implement dom adapter for block tool

ilyamore88:
# Changes description

Changes implement `BlockToolAdapter` class which connects DOM input (for now it works mostly with contenteditable elements) with EditorJSModel.

Adapter listens to `beforeinput` events and dispatches data from the events to the model.
On model update, adapter updates the DOM with relevant changes. We prevent all `beforeinput` events, so all updates to the DOM comes from the model.

`BlockToolAdapter` utilises the `CaretAdapter` to update document selection after a model update.

# CaretAdapter

`CaretAdapter` now updates the model on selection change and, vice versa, updates the DOM on the model updates.

However, current implementation works only with one attached input, we would beed to re-implement it to work with the whole EditorJS document

# Model updates

## Caret

Class represents user caret in the EditorJSModel

Public signature

```ts
class Caret  extends EventBus {
  /** Caret identifier */
  id: number;
  
  /** Caret`s current index */
  index: Index;

  /** 
   * Updates caret index, dispatches an event on update
   **/
  public update(index: Index | null);
}
```

## CaretManager

Class is a part of EditorJSModel which store registry of user carets.

Public signature

```ts
class CaretManager extends EventBus {
  /**
   * Creates a caret instance and puts it inthe registry
   **/
  public createCaret(index?: Index);

  /**
   * Updates a caret instance in the registry
   **/
  public updateCaret(caret: Caret);

  /**
   * Removes a caret instance from the registry
   **/
  public removeCaret(caret: Caret);
}
```

Each method dispatches corresponding event.

All `CaretManagement` methods and events are exposed to the `EditorJSModel` public interface. So carets could be updated from external code (like dom-adapters)


neSpecc (commented):
Please, fill the PR description will main idea and details, description of changes

packages/dom-adapters/src/BlockToolAdapter/index.ts
```diff
@@ -0,0 +1,196 @@
...
+
+  /**
+   * Attaches input to the model using key
+   * It handles beforeinput events and updates model data
+   *
+   * @param key - data key to attach input to
+   * @param input - input element
+   */
+  public attachInput(key: DataKey, input: HTMLElement): void {
+    const caretAdapter = new CaretAdapter(input, this.#model, this.#blockIndex, key);
```
neSpecc, line 51: probably adapter creation should be placed after `tagName` check

packages/dom-adapters/src/BlockToolAdapter/index.ts
```diff
@@ -0,0 +1,196 @@
...
+
+    const inputTag = input.tagName as NativeInput;
+
+    if (![NativeInput.Textarea, NativeInput.Input].includes(inputTag) && !input.isContentEditable) {
```
neSpecc, line 55: not all inputs are text-editable

packages/dom-adapters/src/BlockToolAdapter/index.ts
```diff
@@ -0,0 +1,196 @@
...
+    const range = targetRanges[0];
+
+    const start = getAbsoluteRangeOffset(input, range.startContainer, range.startOffset);
+    const end = getAbsoluteRangeOffset(input, range.endContainer, range.endOffset);
+
+    switch (inputType) {
+      case InputType.InsertReplacementText: {
+        this.#model.removeText(this.#blockIndex, key, start, end);
+
+        const data = event.dataTransfer!.getData('text/plain');
```
neSpecc, line 90: why dataTransfer check is ignored?

packages/dom-adapters/src/BlockToolAdapter/index.ts
```diff
@@ -0,0 +1,196 @@
...
+      return;
+    }
+
+    const [rangeIndex, dataIndex, blockIndex] = event.detail.index;
+
+    if (blockIndex !== this.#blockIndex) {
+      return;
+    }
+
+    if (dataIndex !== `data@${key}`) {
```
neSpecc, line 153: its better to move the data key composing to some helper

packages/dom-adapters/src/BlockToolAdapter/index.ts
```diff
@@ -0,0 +1,196 @@
...
+      return;
+    }
+
+    if (!Array.isArray(rangeIndex)) {
+      return;
```
neSpecc, line 158: it looks like exception

packages/model/src/caret/CaretManager.ts
```diff
@@ -0,0 +1,57 @@
...
+   * EditorJSModel instance
+   */
+  #model: EditorJSModel;
+
+  /**
+   * CaretManager constructor
+   *
+   * @param model - EditorJSModel instance
+   */
+  constructor(model: EditorJSModel) {
```
neSpecc, line 25: Does CaretManager should depend on the Model?

packages/model/src/utils/EventBus/events/CaretUpdatedEvent.ts
neSpecc, line 1: seems like it is not a util

packages/model/src/utils/index.ts
ilyamore88, line 1: You can delete this file for now. EventBus is outside of the utils folder.

packages/model/src/EventBus/composeIndex.ts
ilyamore88, line 1: Rename the file to composeDataIndex.

ilyamore88: Is this function a part of EventBus? Maybe move it to utils?

packages/model/src/CaretManagement/Caret/Caret.ts
```diff
@@ -0,0 +1,88 @@
...
+  public update(index: Index): void {
+    this.#index = index;
+
+    this.dispatchEvent(new CaretUpdatedEvent(this));
+  }
+
+  /**
+   * Serializes caret to JSON
+   */
+  public toJSON(): CaretSerialized {
```
ilyamore88, line 82: Why not .serialized getter? As in nodes.

packages/model/src/CaretManagement/Caret/Caret.ts
```diff
@@ -0,0 +1,88 @@
...
+}
+
+/**
+ * Caret is responsible for storing caret index
+ */
+export class Caret extends EventBus {
+  /**
+   * Caret index
+   */
+  #index: Index | null = null;
```
ilyamore88, line 33: When caret index can be null? I think that in this case, we should delete the Caret instance because the index is null and this caret doesn't exist.

packages/model/src/EventBus/composeIndex.ts
gohabereg: Indexes are part of EventBus now, I've already moved it out of utils. Let's maybe do restructuring in a separate PR after a discussion

packages/model/src/CaretManagement/Caret/Caret.ts
gohabereg: In case we'd use `JSON.stringify` on the caret object to send over the network this method would be natively used

packages/model/src/CaretManagement/Caret/Caret.ts
gohabereg: I'm not sure for that at the moment. If user unfocuses the editor should we remove the caret or should we just set it to null. Also it could be null on creation

packages/dom-adapters/src/BlockToolAdapter/index.ts
neSpecc: this case still not clear. Add jsdoc please describing how it can happen

packages/playground/src/App.vue
```diff
@@ -1,10 +1,28 @@
...
+  } ],
+});
+const document = ref(new EditorDocument(model.serialized));
+
+const serialized = ref(model.serialized);
+
+model.addEventListener(EventType.Changed, () => {
+  serialized.value = model.serialized;
+  document.value = new EditorDocument(model.serialized);
+});
```
neSpecc, line 25: look strange, that you're getting information using existing `model.serialized`, but not from the Event you're listening. Is it the best way?

packages/dom-adapters/src/BlockToolAdapter/index.ts
gohabereg: Hm, I have removed it, not sure why it's still in the PR

packages/playground/src/App.vue
gohabereg: Event has only the bit of info which was changed (e.g. some data or block) so we need to have the full object from the model

neSpecc (approved):

gohabereg (approved):

PR #68, opened 2024-01-24
feat: improve `BlockToolAdapter`

himself65:
## Checklist
- [x] Native inputs handling
- Input element
- Textarea element
- Handle non-text input elements
- [x] Contenteditable element handling
- [x] improve playground

- supported `backspace` and `delete` for native inputs
Left todo for native inputs (support all deletion cases manually)

### DIFF
--- a/packages/dom-adapters/package.json
+++ b/packages/dom-adapters/package.json
@@ -6,8 +6,8 @@
   "types": "dist/index.d.ts",
   "type": "module",
   "scripts": {
-    "build": "tsc --build tsconfig.build.json",
-    "dev": "yarn build --watch",
+    "build": "tsc --project tsconfig.build.json",
+    "dev": "tsc --project tsconfig.build.json --watch",
     "lint": "eslint ./src",
     "lint:ci": "yarn lint --max-warnings 0",
     "lint:fix": "yarn lint --fix",
--- a/packages/dom-adapters/package.json
+++ b/packages/dom-adapters/package.json
@@ -33,6 +33,7 @@
     "typescript": "^5.2.2"
   },
   "dependencies": {
+    "@editorjs/dom": "^1.0.0",
     "@editorjs/model": "workspace:^"
   }
 }
--- a/packages/dom-adapters/package.json
+++ b/packages/dom-adapters/package.json
@@ -6,8 +6,8 @@
   "types": "dist/index.d.ts",
   "type": "module",
   "scripts": {
-    "build": "tsc --project tsconfig.build.json",
-    "dev": "tsc --project tsconfig.build.json --watch",
+    "build": "tsc --build tsconfig.build.json",
+    "dev": "yarn build --watch",
     "lint": "eslint ./src",
     "lint:ci": "yarn lint --max-warnings 0",
     "lint:fix": "yarn lint --fix",
--- a/packages/dom-adapters/src/BlockToolAdapter/index.ts
+++ b/packages/dom-adapters/src/BlockToolAdapter/index.ts
@@ -238,24 +238,10 @@ export class BlockToolAdapter {
         break;
       }
       case EventAction.Removed: {
-        if (
-          start === end &&
-          start > 0
-        ) {
-          currentElement.value = currentElement.value.slice(0, start - 1) +
-            currentElement.value.slice(end);
-          if (start === 0) {
-            currentElement.setSelectionRange(start, start);
-          } else {
-            currentElement.setSelectionRange(start - 1, start - 1);
-          }
-          break;
-        } else {
-          currentElement.value = currentElement.value.slice(0, start) +
-            currentElement.value.slice(end);
-          currentElement.setSelectionRange(start, start);
-          break;
-        }
+        currentElement.value = currentElement.value.slice(0, start) +
+          currentElement.value.slice(end);
+        currentElement.setSelectionRange(start, start);
+        break;
       }
     }
   };
--- a/packages/dom-adapters/src/BlockToolAdapter/index.ts
+++ b/packages/dom-adapters/src/BlockToolAdapter/index.ts
@@ -105,7 +105,7 @@ export class BlockToolAdapter {
           break;
         }
         case InputType.DeleteContentBackward: {
-          this.#model.removeText(this.#blockIndex, key, start, end);
+          this.#model.removeText(this.#blockIndex, key, start - 1, end);
           break;
         }
       }
@@ -224,6 +224,7 @@ export class BlockToolAdapter {
           break;
         }
         case EventAction.Removed: {
+          console.log('remove', start, end);
           if (start === end) {
             currentElement.value = currentElement.value.slice(0, start - 1) +
               currentElement.value.slice(end);
--- a/packages/dom-adapters/src/BlockToolAdapter/index.ts
+++ b/packages/dom-adapters/src/BlockToolAdapter/index.ts
@@ -13,6 +13,13 @@ enum NativeInput {
   Input = 'INPUT',
 }
 
+const enum InputMode {
+  Native = 'native',
+  ContentEditable = 'contenteditable',
+}
+
+const NATIVE_INPUT_SET = new Set([NativeInput.Textarea, NativeInput.Input]) as ReadonlySet<NativeInput>;
+
 /**
  * BlockToolAdapter is using inside Block tools to connect browser DOM elements to the model
  * It can handle beforeinput events and update model data
@@ -28,6 +35,11 @@ export class BlockToolAdapter {
    * Index of the block that this adapter is connected to
    */
   #blockIndex: number;
+
+  /**
+   * Input mode
+   */
+  #mode: InputMode = InputMode.Native;
 
   /**
    * BlockToolAdapter constructor
@@ -50,10 +62,11 @@ export class BlockToolAdapter {
   public attachInput(key: DataKey, input: HTMLElement): void {
     const inputTag = input.tagName as NativeInput;
 
-    /**
-     * @todo Filter non-text-editable inputs
-     */
-    if (![NativeInput.Textarea, NativeInput.Input].includes(inputTag) && !input.isContentEditable) {
+    if (NATIVE_INPUT_SET.has(inputTag)) {
+      this.#mode = InputMode.Native;
+    } else if (input.isContentEditable) {
+      this.#mode = InputMode.ContentEditable;
+    } else {
       throw new Error('BlockToolAdapter: input should be either INPUT, TEXTAREA or contenteditable element');
     }
 
@@ -73,7 +86,7 @@ export class BlockToolAdapter {
    * @param input - input element
    * @param key - data key input is attached to
    */
-  #handleBeforeInputEvent(event: InputEvent, input: HTMLElement, key: DataKey): void {
+  #handleBeforeInputEvent = (event: InputEvent, input: HTMLElement, key: DataKey): void => {
     /**
      * We prevent all events to handle them manually via model update
      */
@@ -140,7 +153,7 @@ export class BlockToolAdapter {
 
       default:
     }
-  }
+  };
 
   /**
    * Handles model update events and updates DOM
@@ -150,7 +163,7 @@ export class BlockToolAdapter {
    * @param key - data key input is attached to
    * @param caretAdapter - caret adapter instance
    */
-  #handleModelUpdate(event: ModelEvents, input: HTMLElement, key: DataKey, caretAdapter: CaretAdapter): void {
+  #handleModelUpdate = (event: ModelEvents, input: HTMLElement, key: DataKey, caretAdapter: CaretAdapter): void => {
     if (!(event instanceof TextAddedEvent) && !(event instanceof TextRemovedEvent)) {
       return;
     }
@@ -205,5 +218,5 @@ export class BlockToolAdapter {
     }
 
     input.normalize();
-  }
+  };
 }
--- a/packages/dom-adapters/src/BlockToolAdapter/index.ts
+++ b/packages/dom-adapters/src/BlockToolAdapter/index.ts
@@ -90,89 +90,81 @@ export class BlockToolAdapter {
 
     const isNativeInput = this.#mode === InputMode.Native;
     const inputType = event.inputType as InputType;
+    let start: number;
+    let end: number;
 
-    if (isNativeInput) {
-      const currentElement = input as HTMLInputElement | HTMLTextAreaElement;
-      const start = currentElement.selectionStart as number;
-      const end = currentElement.selectionEnd as number;
-
-      switch (inputType) {
-        case InputType.InsertText: {
-          const data = event.data as string;
-
-          this.#model.insertText(this.#blockIndex, key, data, start);
-
-          break;
-        }
-        case InputType.DeleteContentBackward: {
-          this.#model.removeText(this.#blockIndex, key, start - 1, end);
-          break;
-        }
-      }
-    } else {
+    if (!isNativeInput) {
       const targetRanges = event.getTargetRanges();
       const range = targetRanges[0];
 
-      const start = getAbsoluteRangeOffset(input, range.startContainer,
+      start = getAbsoluteRangeOffset(input, range.startContainer,
         range.startOffset);
-      const end = getAbsoluteRangeOffset(input, range.endContainer,
+      end = getAbsoluteRangeOffset(input, range.endContainer,
         range.endOffset);
+    } else {
+      const currentElement = input as HTMLInputElement | HTMLTextAreaElement;
 
-      switch (inputType) {
-        case InputType.InsertReplacementText:
-        case InputType.InsertFromPaste: {
-          this.#model.removeText(this.#blockIndex, key, start, end);
+      start = currentElement.selectionStart as number;
+      end = currentElement.selectionEnd as number;
+    }
 
-          /**
-           * DataTransfer object is guaranteed to be not null for these types of event for contenteditable elements
-           *
-           * However, it is not guaranteed for INPUT and TEXTAREA elements, so @todo handle this case
-           *
-           * @see https://www.w3.org/TR/input-events-2/#overview
-           */
-          const data = event.dataTransfer!.getData('text/plain');
+    switch (inputType) {
+      case InputType.InsertReplacementText:
+      case InputType.InsertFromPaste: {
+        this.#model.removeText(this.#blockIndex, key, start, end);
 
-          this.#model.insertText(this.#blockIndex, key, data, start);
-
-          break;
-        }
-        case InputType.InsertText:
         /**
-         * @todo Handle composition events
+         * DataTransfer object is guaranteed to be not null for these types of event for contenteditable elements
+         *
+         * However, it is not guaranteed for INPUT and TEXTAREA elements, so @todo handle this case
+         *
+         * @see https://www.w3.org/TR/input-events-2/#overview
          */
-        case InputType.InsertCompositionText: {
-          /**
-           * If start and end aren't equal,
-           * it means that user selected some text and replaced it with new one
-           */
-          if (start !== end) {
-            this.#model.removeText(this.#blockIndex, key, start, end);
-          }
+        const data = event.dataTransfer!.getData('text/plain');
 
-          const data = event.data as string;
+        this.#model.insertText(this.#blockIndex, key, data, start);
 
-          this.#model.insertText(this.#blockIndex, key, data, start);
-          break;
-        }
-
-        case InputType.DeleteContent:
-        case InputType.DeleteContentBackward:
-        case InputType.DeleteContentForward:
-        case InputType.DeleteByCut:
-        case InputType.DeleteByDrag:
-        case InputType.DeleteHardLineBackward:
-        case InputType.DeleteHardLineForward:
-        case InputType.DeleteSoftLineBackward:
-        case InputType.DeleteSoftLineForward:
-        case InputType.DeleteWordBackward:
-        case InputType.DeleteWordForward: {
-          this.#model.removeText(this.#blockIndex, key, start, end);
-
-          break;
-        }
-
-        default:
+        break;
       }
+      case InputType.InsertText:
+      /**
+       * @todo Handle composition events
+       */
+      case InputType.InsertCompositionText: {
+        /**
+         * If start and end aren't equal,
+         * it means that user selected some text and replaced it with new one
+         */
+        if (start !== end) {
+          this.#model.removeText(this.#blockIndex, key, start, end);
+        }
+
+        const data = event.data as string;
+
+        this.#model.insertText(this.#blockIndex, key, data, start);
+        break;
+      }
+
+      case InputType.DeleteContent:
+      case InputType.DeleteContentBackward:
+      case InputType.DeleteContentForward:
+      case InputType.DeleteByCut:
+      case InputType.DeleteByDrag:
+      case InputType.DeleteHardLineBackward:
+      case InputType.DeleteHardLineForward:
+      case InputType.DeleteSoftLineBackward:
+      case InputType.DeleteSoftLineForward:
+      case InputType.DeleteWordBackward:
+      case InputType.DeleteWordForward: {
+        if (isNativeInput && start > 0) {
+          start = start - 1;
+        }
+        this.#model.removeText(this.#blockIndex, key, start, end);
+
+        break;
+      }
+
+      default:
     }
   };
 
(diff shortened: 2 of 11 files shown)


packages/dom-adapters/src/BlockToolAdapter/index.ts
```diff
@@ -171,39 +207,70 @@ export class BlockToolAdapter {
...
 
-    const action = event.detail.action;
+    if (isNativeInput) {
+      const currentElement = input as HTMLInputElement | HTMLTextAreaElement;
+      const start = currentElement.selectionStart!;
+      const end = currentElement.selectionEnd!;
+
+      const action = event.detail.action;
+
+      switch (action) {
```
himself65, line 217: I'm a little worried about this part of logic that doesn't use the `caretAdapter` class. Because onUpdate callback cannot get any information about the update from caretEvent

packages/dom-adapters/src/BlockToolAdapter/index.ts
```diff
@@ -73,74 +82,99 @@ export class BlockToolAdapter {
    * @param input - input element
    * @param key - data key input is attached to
    */
-  #handleBeforeInputEvent(event: InputEvent, input: HTMLElement, key: DataKey): void {
+  #handleBeforeInputEvent = (event: InputEvent, input: HTMLElement, key: DataKey): void => {
```
gohabereg, line 85: If you want to save context here, would be better to use an anonymous arrow function when adding a listener. Arrow function methods are treated as properties, therefore created for each instance (which we expect a lot for a document)

packages/dom-adapters/src/BlockToolAdapter/index.ts
```diff
@@ -73,74 +82,99 @@ export class BlockToolAdapter {
...
     event.preventDefault();
 
+    const isNativeInput = this.#mode === InputMode.Native;
     const inputType = event.inputType as InputType;
-    const targetRanges = event.getTargetRanges();
-    const range = targetRanges[0];
 
-    const start = getAbsoluteRangeOffset(input, range.startContainer, range.startOffset);
-    const end = getAbsoluteRangeOffset(input, range.endContainer, range.endOffset);
+    if (isNativeInput) {
```
gohabereg, line 94: Arguably, would be more maintainable to use the condition in each relevant case rather than creating two switch statements

packages/dom-adapters/src/BlockToolAdapter/index.ts
gohabereg: Agree, would be better if CaretAdapter handles native inputs selection on its own. I guess you can leave a to-do here to consider that for caret adapter re-implementation

packages/dom-adapters/src/BlockToolAdapter/index.ts
```diff
@@ -171,39 +207,70 @@ export class BlockToolAdapter {
...
+          currentElement.value = currentElement.value.slice(0, start) + text + currentElement.value.slice(end);
 
-    const start = rangeIndex[0];
-    const end = rangeIndex[1];
+          currentElement.setSelectionRange(start + text.length, start + text.length);
+          break;
+        }
+        case EventAction.Removed: {
+          if (start === end) {
+            currentElement.value = currentElement.value.slice(0, start - 1) +
```
gohabereg, line 228: Current value could be assigned to a variable before switch statement and reused across the method

packages/dom-adapters/src/utils/index.ts
```diff
@@ -1,2 +1,14 @@
 export * from './getAbsoluteRangeOffset.js';
 export * from './getRelativeIndex.js';
+
+export enum NativeInput {
+  Textarea = 'TEXTAREA',
+  Input = 'INPUT',
```
gohabereg, line 6: There are some non-text input types, but I think we shouldn't worry about them. Just throw an error if a tool tries to attach such input to adapter. Those inputs should be treated as value nodes and handled on a tool side.

What do you think @neSpecc @ilyamore88 ?

neSpecc: Yes. But it should be handled explicitly. Here you can check the input types supporting `selectionStart`: https://html.spec.whatwg.org/multipage/input.html#concept-input-apply:~:text=·-,selectionStart,-·

packages/dom-adapters/src/utils/index.ts
neSpecc, line 1: I'd suggest to leave `index.ts` just as index file and move your utils to `utils/input.ts`

packages/dom-adapters/src/BlockToolAdapter/index.ts
```diff
@@ -29,6 +32,11 @@ export class BlockToolAdapter {
    */
   #blockIndex: number;
 
+  /**
+   * Input mode
+   */
+  #mode: InputMode = InputMode.Native;
```
neSpecc, line 38: Block Tool can have several inputs of different type

packages/dom-adapters/src/BlockToolAdapter/index.ts
```diff
@@ -50,10 +58,11 @@ export class BlockToolAdapter {
   public attachInput(key: DataKey, input: HTMLElement): void {
     const inputTag = input.tagName as NativeInput;
 
-    /**
-     * @todo Filter non-text-editable inputs
-     */
-    if (![NativeInput.Textarea, NativeInput.Input].includes(inputTag) && !input.isContentEditable) {
+    if (Boolean(NATIVE_INPUT_SET.has(inputTag))) {
```
neSpecc, line 61: maybe create `isNativeInput()` util instead of this set for better readability?

packages/dom-adapters/src/BlockToolAdapter/index.ts
```diff
@@ -133,14 +156,17 @@ export class BlockToolAdapter {
       case InputType.DeleteSoftLineForward:
       case InputType.DeleteWordBackward:
       case InputType.DeleteWordForward: {
+        if (isNativeInput && start > 0) {
+          start = start - 1;
+        }
```
neSpecc, line 161: add a comment here please

packages/playground/src/components/Input.vue
```diff
@@ -20,6 +20,16 @@ const props = defineProps<{
   model: EditorJSModel;
 }>();
 
+// three types of inputs:
+// ?type=contenteditable
+// ?type=input
+// ?type=textarea
+let type = (new URL(window.location.href).searchParams.get('type') ?? 'contenteditable') as 'contenteditable' | 'input' | 'textarea';
```
neSpecc, line 27: you can use vue component props instead

himself65: I think it will be simpler if I read it directly.

himself65 (commented):
for html textarea, should we create a new paragraph or something when user press `Enter`?

gohabereg:
> for html textarea, should we create a new paragraph or something when user press `Enter`?

No, we'd like to preserve browser behaviour there. If some other behaviour is desired by a tool it should be implemented by the tool

himself65:
> We'd like to preserve browser behavior there.

Is it a good idea that keep the default behavior?

image

Benefit is we don't need to handle all cases for cross-browser manually, but I think there would be some issue when bringing the collaborative feature in

packages/playground/src/components/Input.vue
neSpecc: I still think that props are more suitable here. Query params are related to the page itself, when props are related to the component instance. Maybe we'll need several Inputs of different types on the Playground.

Props will be quite simpler:

**Input.vue**

```ts
defineProps<{
  type: 'contenteditable' | 'input' | 'textarea'
}>()
```

**Playground.vue**

```html
<Input type="contenteditable" />
<Input type="input" />
```

himself65: For now, we don't support multiple input elements on the same page, and cannot auto-unmount input elements.

packages/dom-adapters/package.json
```diff
@@ -6,8 +6,8 @@
   "types": "dist/index.d.ts",
   "type": "module",
   "scripts": {
-    "build": "tsc --build tsconfig.build.json",
-    "dev": "yarn build --watch",
+    "build": "tsc --project tsconfig.build.json",
+    "dev": "tsc --project tsconfig.build.json --watch",
```
neSpecc, line 10: Please, describe these changes in a PR description

packages/dom-adapters/src/BlockToolAdapter/index.ts
```diff
@@ -133,14 +153,24 @@ export class BlockToolAdapter {
...
+        if (nativeInput && start > 0 && start === end) {
+          // for native input elements, we need to handle backspace manually
+          start = start - 1;
+        }
+        switch (inputType) {
+          case InputType.DeleteSoftLineBackward: {
+            start = 0;
+            break;
+          }
+        }
```
neSpecc, line 165: this switch seems redundant since you're already in this case

packages/dom-adapters/src/BlockToolAdapter/index.ts
```diff
@@ -133,14 +153,24 @@ export class BlockToolAdapter {
       case InputType.DeleteSoftLineForward:
       case InputType.DeleteWordBackward:
       case InputType.DeleteWordForward: {
+        if (nativeInput && start > 0 && start === end) {
+          // for native input elements, we need to handle backspace manually
+          start = start - 1;
+        }
```
neSpecc, line 159: I'd suggest not to mix different implementation in a single code. Lets move them to separate methods, like 

`handleDeleteInNativeInput()`
`handleDeleteInContentEditable()`

packages/dom-adapters/src/BlockToolAdapter/index.ts
```diff
@@ -171,39 +203,83 @@ export class BlockToolAdapter {
       return;
     }
 
-    const action = event.detail.action;
+    if (nativeInput) {
+      const currentElement = input as HTMLInputElement | HTMLTextAreaElement;
```
neSpecc, line 207: same suggestion, let's move logic to separate methods

packages/dom-adapters/src/BlockToolAdapter/index.ts
```diff
@@ -171,39 +203,83 @@ export class BlockToolAdapter {
...
+      const [ [start, end] ] = event.detail.index;
+
+      // todo:
+      //  caretAdapter cannot handle native input
+      //  because it doesn't have the content of the input
```
neSpecc, line 212: Please, use block-style comments and JSDoc-style todos as we use in a remaining code
```suggestion
      /** 
       * @todo caretAdapter cannot handle native because it doesn't have the content of the input
       */
```

packages/dom-adapters/src/caret/utils/useSelectionChange.ts
```diff
@@ -52,6 +52,11 @@ export const useSelectionChange = createSingleton(() => {
       return false;
     }
 
+    if (selection.rangeCount <= 0) {
+      // special case for native input elements
+      return false;
+    }
```
neSpecc, line 58: 1) Please use block-style comment
2) Put it above to related code 
3) Explain more details: why it was added, which case you're mentioning

packages/dom-adapters/src/caret/CaretAdapter.ts
```diff
@@ -187,17 +191,24 @@ export class CaretAdapter extends EventTarget {
...
-    const end = getBoundaryPointByAbsoluteOffset(this.#input!, textIndex[1]);
+    const input: HTMLElement = this.#input!;
 
-    const selection = document.getSelection()!;
-    const range = new Range();
+    if (isNativeInput(input)) {
+      // skip
+    } else {
+      const start = getBoundaryPointByAbsoluteOffset(input,
+        textIndex[0]);
```
neSpecc, line 200: bad linebreak

packages/dom-adapters/src/caret/CaretAdapter.ts
```diff
@@ -187,17 +191,24 @@ export class CaretAdapter extends EventTarget {
...
 
-    const start = getBoundaryPointByAbsoluteOffset(this.#input!, textIndex[0]);
-    const end = getBoundaryPointByAbsoluteOffset(this.#input!, textIndex[1]);
+    const input: HTMLElement = this.#input!;
 
-    const selection = document.getSelection()!;
-    const range = new Range();
+    if (isNativeInput(input)) {
+      // skip
+    } else {
```
neSpecc, line 198:
```suggestion
    /**
     * ...
     */
    if (isNativeInput(input)) {
      return;
    } 
```

packages/dom-adapters/src/BlockToolAdapter/index.ts
```diff
@@ -73,18 +79,31 @@ export class BlockToolAdapter {
...
    * @param key - data key input is attached to
    */
-  #handleBeforeInputEvent(event: InputEvent, input: HTMLElement, key: DataKey): void {
+  #handleBeforeInputEvent = (event: InputEvent, input: HTMLElement, key: DataKey): void => {
     /**
      * We prevent all events to handle them manually via model update
      */
     event.preventDefault();
 
+    const nativeInput = this.#mode === InputMode.Native;
```
neSpecc, line 88: it is the main problem right now, BlockToolAdapter should support several inputs

neSpecc (changes requested):

packages/dom-adapters/src/BlockToolAdapter/index.ts
```diff
@@ -48,22 +50,68 @@ export class BlockToolAdapter {
...
+    const targetRanges = event.getTargetRanges();
+    const range = targetRanges[0];
+
+    start = getAbsoluteRangeOffset(input, range.startContainer, range.startOffset);
+    end = getAbsoluteRangeOffset(input, range.endContainer, range.endOffset);
+
+    if (start > 0 && start === end) {
+      start = start - 1;
+    }
+    switch (inputType) {
```
gohabereg, line 106: Please add some comments for the if statement, it's not clear why it is needed

gohabereg, line 106:
```suggestion
    if (start > 0 && start === end) {
      start = start - 1;
    }

    switch (inputType) {
```

packages/dom-adapters/src/BlockToolAdapter/index.ts
```diff
@@ -48,22 +50,68 @@ export class BlockToolAdapter {
...
+   *
+   * @param event - beforeinput event
+   * @param input - input element
+   * @param key - data key input is attached to
+   * @private
+   */
+  #handleDeleteInNativeInput = (event: InputEvent, input: HTMLInputElement | HTMLTextAreaElement, key: DataKey): void => {
+    const inputType = event.inputType as InputType;
+    let start = input.selectionStart as number;
+    const end = input.selectionEnd as number;
```
gohabereg, line 71: If value could be null we need to handle that case to avoid runtime errors

packages/dom-adapters/src/BlockToolAdapter/index.ts
```diff
@@ -48,22 +50,68 @@ export class BlockToolAdapter {
...
+
+    if (start > 0 && start === end) {
+      start = start - 1;
+    }
+    switch (inputType) {
```
gohabereg, line 76: Please add a comment why this is required
```suggestion
    if (start > 0 && start === end) {
      start = start - 1;
    }

   switch (inputType) {
```

gohabereg: Also in case this is a forward delete, we'd need to increase `end` by one

packages/dom-adapters/src/BlockToolAdapter/index.ts
```diff
@@ -48,22 +50,68 @@ export class BlockToolAdapter {
...
+      case InputType.DeleteSoftLineBackward: {
```
gohabereg, line 77: Do we need to handle forward delete the same way here?

packages/dom-adapters/src/BlockToolAdapter/index.ts
```diff
@@ -133,40 +195,82 @@ export class BlockToolAdapter {
...
-  /**
-   * Handles model update events and updates DOM
-   *
-   * @param event - model update event
-   * @param input - attched input element
-   * @param key - data key input is attached to
-   * @param caretAdapter - caret adapter instance
-   */
-  #handleModelUpdate(event: ModelEvents, input: HTMLElement, key: DataKey, caretAdapter: CaretAdapter): void {
+  #handleModelUpdateNative = (event: ModelEvents, input: HTMLInputElement | HTMLTextAreaElement, key: DataKey): void => {
```
gohabereg, line 210:
```suggestion
  #handleModelUpdateForNativeInput = (event: ModelEvents, input: HTMLInputElement | HTMLTextAreaElement, key: DataKey): void => {
```

packages/dom-adapters/src/BlockToolAdapter/index.ts
```diff
@@ -133,40 +195,82 @@ export class BlockToolAdapter {
...
+          currentElement.value = currentElement.value.slice(0, start) +
+            currentElement.value.slice(end);
+          currentElement.setSelectionRange(start, start);
+          break;
+        }
+      }
+    }
+  };
+
+  #handleModelUpdateContentEditable = (event: ModelEvents, input: HTMLElement, key: DataKey, caretAdapter: CaretAdapter): void => {
```
gohabereg, line 263:
```suggestion
  #handleModelUpdateForContentEditableElement = (event: ModelEvents, input: HTMLElement, key: DataKey, caretAdapter: CaretAdapter): void => {
```

packages/dom-adapters/src/BlockToolAdapter/index.ts
```diff
@@ -133,40 +195,82 @@ export class BlockToolAdapter {
...
-   * @param key - data key input is attached to
-   * @param caretAdapter - caret adapter instance
-   */
-  #handleModelUpdate(event: ModelEvents, input: HTMLElement, key: DataKey, caretAdapter: CaretAdapter): void {
+  #handleModelUpdateNative = (event: ModelEvents, input: HTMLInputElement | HTMLTextAreaElement, key: DataKey): void => {
+    if (!(event instanceof TextAddedEvent) && !(event instanceof TextRemovedEvent)) {
+      return;
+    }
+
+    const [, dataIndex, blockIndex] = event.detail.index;
```
gohabereg, line 215: We use `dataKey` in other places, let's keep it for consistency

```suggestion
    const [, dataKey, blockIndex] = event.detail.index;
```

packages/dom-adapters/src/BlockToolAdapter/index.ts
```diff
@@ -133,40 +195,82 @@ export class BlockToolAdapter {
...
+
+    if (blockIndex !== this.#blockIndex) {
+      return;
+    }
+
+    if (dataIndex !== composeDataIndex(key)) {
+      return;
+    }
+
+    const currentElement = input as HTMLInputElement | HTMLTextAreaElement;
```
gohabereg, line 225: Seems to be unnecessary cast here, `input` already has `HTMLInputElement | HTMLTextAreaElement` type

packages/dom-adapters/src/BlockToolAdapter/index.ts
```diff
@@ -133,40 +195,82 @@ export class BlockToolAdapter {
...
+        const prevValue = currentElement.value;
+
+        currentElement.value = prevValue.slice(0, start) + text + prevValue.slice(end - 1);
+
+        currentElement.setSelectionRange(start + text.length, start + text.length);
+        break;
+      }
+      case EventAction.Removed: {
+        if (
+          start === end &&
```
gohabereg, line 242: I don't think we can get this case from the model update. If something is deleted, start would not be equal to end

packages/dom-adapters/src/BlockToolAdapter/index.ts
```diff
@@ -133,40 +195,82 @@ export class BlockToolAdapter {
...
+          start > 0
+        ) {
+          currentElement.value = currentElement.value.slice(0, start - 1) +
+            currentElement.value.slice(end);
+          if (start === 0) {
```
gohabereg, line 247: In the parent statement you check if `start > 0`, so this condition will always be false

packages/dom-adapters/src/BlockToolAdapter/index.ts
```diff
@@ -176,7 +280,8 @@ export class BlockToolAdapter {
     const start = rangeIndex[0];
     const end = rangeIndex[1];
 
-    const [startNode, startOffset] = getBoundaryPointByAbsoluteOffset(input, start);
+    const [startNode, startOffset] = getBoundaryPointByAbsoluteOffset(input,
+      start);
```
gohabereg, line 284: Seems to be unnecessary change here

packages/dom-adapters/src/BlockToolAdapter/index.ts
```diff
@@ -133,40 +195,82 @@ export class BlockToolAdapter {
...
+  };
+
+  #handleModelUpdateContentEditable = (event: ModelEvents, input: HTMLElement, key: DataKey, caretAdapter: CaretAdapter): void => {
     if (!(event instanceof TextAddedEvent) && !(event instanceof TextRemovedEvent)) {
       return;
     }
 
     const [rangeIndex, dataIndex, blockIndex] = event.detail.index;
 
-    /**
```
gohabereg, line 160: Could you please revert the comments

packages/dom-adapters/src/utils/index.ts
gohabereg: Still relevant

gohabereg (commented):
Not covered cases:
- Forward delete of a single character in native `INPUT` element
- Forward soft line delete in native `INPUT` element
- `insertFromDrop` event type in native `INPUT` element
- Likely `hardLineDelete` event type doesn't work at all
- `deleteWordBackward/Forward` event type doesn't work for native `INPUT` element
- The same problems for native `TEXTAREA` element
- Buggy behaviour for `deleteWordForward` for contenteditable element if caret is at the end of the line
- `deleteSoftLineForward` event type is not handled for contenteditable elements
- `deleteHardLineBackward/Forward` event types are not handled for contenteditable elements
- `insertFromDrop` event type is not handled at all
- `deleteEntireSoftLine` event type is not handled at all

I think we can merge this PR first and then work on each case in a separate PR

packages/dom-adapters/src/utils/index.ts
e11sy: image
handled

packages/dom-adapters/src/BlockToolAdapter/index.ts
```diff
@@ -73,18 +130,31 @@ export class BlockToolAdapter {
    * @param input - input element
    * @param key - data key input is attached to
    */
-  #handleBeforeInputEvent(event: InputEvent, input: HTMLElement, key: DataKey): void {
+  #handleBeforeInputEvent = (event: InputEvent, input: HTMLElement, key: DataKey): void => {
```
gohabereg, line 133: Looks like this could be a regular method, not an arrow function

packages/dom-adapters/src/BlockToolAdapter/index.ts
```diff
@@ -73,18 +130,31 @@ export class BlockToolAdapter {
...
     /**
      * We prevent all events to handle them manually via model update
      */
     event.preventDefault();
 
+    const nativeInput = isNativeInput(input);
```
gohabereg, line 139: Better to call such variables with `is`, as `isNativeInput` is already declared, you could name it smth like
```suggestion
    const isInputNative = isNativeInput(input);
```

packages/dom-adapters/src/BlockToolAdapter/index.ts
```diff
@@ -205,5 +310,23 @@ export class BlockToolAdapter {
...
+  /**
+   * Handles model update events and updates DOM
+   *
+   * @param event - model update event
+   * @param input - attched input element
+   * @param key - data key input is attached to
+   * @param caretAdapter - caret adapter instance
+   */
+  #handleModelUpdate = (event: ModelEvents, input: HTMLElement, key: DataKey, caretAdapter: CaretAdapter): void => {
+    const nativeInput = isNativeInput(input);
```
gohabereg, line 324: Same here

packages/dom-adapters/src/caret/utils/useSelectionChange.ts
```diff
@@ -48,7 +48,14 @@ export const useSelectionChange = createSingleton(() => {
    * @param input - input to check
    */
   function isSelectionRelatedToInput(selection: Selection | null, input: InputWithCaret): boolean {
-    if (!selection) {
+    if (selection === null) {
+      return false;
+    }
+
+    /**
+     * Check, that selection range exists in native inputs
```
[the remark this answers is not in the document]
e11sy: this check turned out to be redundant

gohabereg (approved):

packages/playground/src/components/Input.vue
neSpecc, line 1: add style `border: 0` please to remove default browser border
image

neSpecc (approved):

PR #69, opened 2024-01-24
CaretAdapter reimplementation

gohabereg:
CaretAdapter should be global in scope of one Document.

The new data flow is the following:
1. BlockToolAdapter receives CaretAdapter instance as a constructor parameter
2. When input is attached to a BlockToolAdapter instance, it is attached to the CaretAdapter as well
3. CaretAdapter keeps track of all inputs by their `DataIndex` (block index + data property index)
4. When selection is changed inside the Editor container, CaretAdapter receives an update and updates the model
5. When model is updated, CaretAdapter receives an update, looks for related input and sets the selection there

ToDo:
- [x] Refactoring + tests
- [x] Native inputs support
- [x] Think of better hash function rather than just `JSON.stringify`, maybe create a separate class for operation with indexes
- [ ] Caret should not be updated if a model update comes from the remote user — think of how to distinct own and others updates


packages/model/src/EventBus/types/indexing.ts
```diff
@@ -29,7 +29,7 @@ export type TextRange = [number, number];
 /**
  * Generic type for BlockNode elements
```
neSpecc, line 30: Seems like jsdoc is irrelevant, please update

packages/model/src/CaretManagement/Caret/Caret.ts
```diff
@@ -71,6 +71,10 @@ export class Caret extends EventBus {
    * @param index - new caret index
    */
   public update(index: Index): void {
+    if (JSON.stringify(this.#index) === JSON.stringify(index)) {
```
neSpecc, line 74: please add jsdocs for such statements

packages/playground/src/components/Input.vue
```diff
@@ -21,22 +22,33 @@ const props = defineProps<{
...
+    blockToolAdapter.attachInput(createDataKey('text1'), input1.value);
 
     props.model.addEventListener(EventType.CaretManagerUpdated, (evt: CaretManagerCaretUpdatedEvent) => {
-      index.value = (evt.detail.index as TextIndex)[0];
+      index.value = evt.detail.index;
     });
   }
+
+  if (input2.value !== null) {
+    blockToolAdapter.attachInput(createDataKey('text2'), input2.value);
```
neSpecc, line 37: can we get rid of call of "createDataKey" for tools developers? It would be more friendly to just pass string

packages/dom-adapters/src/BlockToolAdapter/index.ts
```diff
@@ -189,7 +190,7 @@ export class BlockToolAdapter {
 
         range.insertNode(textNode);
 
-        caretAdapter.updateIndex([start + text.length, start + text.length]);
+        this.#caretAdapter.updateIndex([[start + text.length, start + text.length], composeDataIndex(key), this.#blockIndex]);
```
neSpecc, line 193: its hard to understand when all params passed as array. I'd suggest creating private (or utility) method like `composeInputCaretIndex(start, end, key, blockIndex)`

packages/dom-adapters/src/CaretAdapter/index.ts
```diff
@@ -0,0 +1,128 @@
+import { useSelectionChange } from '../caret/utils/useSelectionChange.js';
+import type { Caret, EditorJSModel, DataIndex, TextIndex, CaretManagerEvents, Index } from '@editorjs/model';
+import { getAbsoluteRangeOffset, getBoundaryPointByAbsoluteOffset } from '../utils/index.js';
+import type { TextRange } from '@editorjs/model';
+import { EventType } from '@editorjs/model';
+
+function hashIndex(index: Readonly<Index>): string {
```
neSpecc, line 7: its better to move them to utils

packages/dom-adapters/src/CaretAdapter/index.ts
```diff
@@ -0,0 +1,128 @@
...
+   *
+   * @param container
+   * @param model
+   */
+  constructor(container: HTMLElement, model: EditorJSModel) {
+    super();
+
+    this.#model = model;
+    this.#container = container;
+    this.#userCaret = this.#model.createCaret();
```
neSpecc, line 42: probably some user identifier should be passed from somewhere or generated here. Since there can be several user carets

packages/dom-adapters/src/CaretAdapter/index.ts
```diff
@@ -0,0 +1,128 @@
...
+
+    const { on } = useSelectionChange();
+
+    on(container, (selection) => this.#onSelectionChange(selection), this);
```
neSpecc, line 46: to not forget to call "off" somewhere

neSpecc (commented):
Please add well-descriptive jsdocs — it's important because all this stuff became more complex every update

packages/dom-adapters/src/CaretAdapter/index.ts
```diff
@@ -0,0 +1,143 @@
...
+  /**
+   *
+   * @param selection
+   */
+  #onSelectionChange(selection: Selection | null): void {
+    if (!selection) {
+      return;
+    }
+
+    const activeElement = document.activeElement;
```
neSpecc, line 69: there can be blocks common wrapper if it will be content editable

packages/dom-adapters/src/CaretAdapter/index.ts
```diff
@@ -0,0 +1,194 @@
...
+
+      /**
+       * For now we handle only first found input
+       */
+      break;
+    }
+  }
+
+  /**
+   * Model updates handler
```
neSpecc, line 138: Describe what it does

packages/dom-adapters/src/CaretAdapter/index.ts
```diff
@@ -0,0 +1,194 @@
...
+    const caretId = event.detail.id;
+
+    if (caretId !== this.#userCaret.id) {
+      return;
+    }
+
+    const builder = new IndexBuilder();
+
+    /**
+     * We need to remove text range from index to find related input by serialized index
```
neSpecc, line 164: > to find related input by serialized index

sounds unclear

packages/dom-adapters/src/CaretAdapter/index.ts
```diff
@@ -0,0 +1,194 @@
...
+        builder.from(index).addTextRange(textRange);
+
+        this.updateIndex(builder.build());
+
+        /**
+         * For now we handle only first found input
+         */
+        break;
+      }
+
```
neSpecc, line 113: the page freezes and dies when I focus contenteditable input in Safari

neSpecc (commented):
backward mouse range selection is not working

packages/dom-adapters/src/CaretAdapter/index.ts
gohabereg: Fixed

packages/dom-adapters/src/BlockToolAdapter/index.ts
```diff
@@ -228,7 +238,11 @@ export class BlockToolAdapter {
       return;
     }
 
-    const [, dataKey, blockIndex] = event.detail.index;
+    const { textRange, dataKey, blockIndex }  = event.detail.index;
```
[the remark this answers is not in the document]
gohabereg: Builder should have a new Index instance inside, so it can't have static methods

neSpecc (approved):

PR #70, opened 2024-02-03
fix(build): use ts references, add hack for esbuild

ilyamore88:
## Typescript references
Now we're using the `--build` flag to build typescript projects. This flag supports project references. Because of that, we don't need to use Yarn Workspaces on CI to build package dependencies topologically. TSC is responsible for that.

## Vite and ESBuild 🤡
ESBuild doesn't support all tsconfig fields. See https://esbuild.github.io/content-types/#tsconfig-json. I've fixed this problem by creating `build:dependencies` script in the playground package. This script builds all workspace dependencies.

### DIFF
--- a/.github/actions/build/action.yml
+++ b/.github/actions/build/action.yml
@@ -16,4 +16,4 @@ runs:
 
     - name: Build the package with dependencies
       shell: bash
-      run: yarn workspaces foreach -Rpt --from '${{ inputs.package-name }}' run build
+      run: yarn workspace ${{ inputs.package-name }} run build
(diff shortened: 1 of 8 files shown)


neSpecc:
Which problem was resolved by this PR?

ilyamore88:
Hi, @neSpecc. I've described it in the description. Previously, tsc didn't build packages from references. For example, if you try to build dom-adapters without building the model, you get an error. Now, if you build dom-adapters, tsc looks to their references and builds it first. Also, it improves the dev mode. If I run `yarn dev` (tsc build with --watch flag) in the dom-adapters package and change something in the model, tsc rebuilds the model and dom-adapters because it understands the references field now.

gohabereg (approved):

neSpecc (approved):

PR #71, opened 2024-03-10
feat(e2e-testing): setup playwright

ilyamore88:
## ToDo
- [x] Setup base Playwright instance;
- [x] Add CI job for Playwright running;
- [x] Add basic test cases for Playground.


gohabereg (approved):

neSpecc (commented):
We don't need to test playground, we should test dom-adapters. So tests should be setup for the Dom-adapters package

PR #74, opened 2024-08-27
feat(dom-adapters): basic inline tool adapter implementation

e11sy:
### structure:
- `Editor` (playground simulation)
- `InlineTool` instanse (one for editor, for each inline tool)
- `InlineToolbar` toolbar that is rendered with inline tools available inside block
  - Business logic : (`InlineToolbar` class) 
  - UI : `Toolbar` component 
- `InlineToolAdapter` - adapter,  that connects model and editor
- `model` stores data

NOTE: 
- tool should determine it's state for selection

## Progress
- [x] Implement playground simulation
- [ ] `InlineToolbar`
  - [ ] Business logic
    - [x] Subscribe on selection change
    - [ ] Validate inline tools available in current block
  - [ ] UI 
    - [x] Simple realisation
    - [ ] Support additional data for inline tools (e.g. link tool requires link)
- [x] `InlineToolAdapter`
  - [x]  Get `action` and `range` from tool instance
  - [x] Subscribe on model changes
- [ ] Model support `intersectType` for inline tools
``` typescript
enum IntersectType {
  Extend = 'extend',           // if two fragments of one tool intersect - merge into one fragment
  Replace = 'replace',         // if two new fragment intersect with existing - remove existing with new one
  Separate = 'separate'        // if two fragments of one tool intersect - treat them as two different fragments
}
```
- [ ] Tests
image

## Next
- Validate inline tools that are available in current block
- Support additional data for inline tools
- Support `IntersectType` for inline tools
- Add ability to unformat inline tool in dom (that should do tool with adapter)
- Caret should reset state of selection if no selection exists
- Cover with tests


packages/dom-adapters/src/InlineToolAdapter/index.ts
```diff
@@ -0,0 +1,144 @@
...
+
+      /**
+       * Render inline tool for current range
+       */
+      if (selection) {
+        const range = selection.getRangeAt(0);
+
+        const inlineElement = tool.create();
+
+        range.surroundContents(inlineElement);
```
gohabereg, line 82: That would throw an exception in some cases, see https://developer.mozilla.org/en-US/docs/Web/API/Range/surroundContents

packages/dom-adapters/src/InlineToolAdapter/index.ts
```diff
@@ -0,0 +1,144 @@
...
+  public applyFormat(toolName: InlineToolName, data: InlineToolData, intersectType: IntersectType): void {
+    const index = this.#caretAdapter.userCaretIndex;
+
+    if (index === null) {
+      throw new Error('InlineToolAdapter: caret index is outside of the input');
+    }
+
+    const textRange = index.textRange!;
+    const blockIndex = index.blockIndex!;
+    const dataKey = index.dataKey!;
```
gohabereg, line 121: Non-null assertion could be dangerous here, would be better to check each index part

packages/dom-adapters/src/InlineToolAdapter/index.ts
```diff
@@ -0,0 +1,144 @@
...
+
+    const tool = this.#tools.get(toolName);
+
+    if (!tool) {
+      throw new Error(`InlineToolAdapter: tool ${toolName} is not attached`);
+    }
+
+    const fragments = this.#model.getFragments(blockIndex, dataKey, ...textRange, toolName);
+
+    const { action, range } = tool.getAction(textRange, fragments, intersectType, data);
```
gohabereg, line 131: Tool will know it's intersect type by itself, no need to pass it here. It should be passed to the model when we'll implement different intersections support

packages/dom-adapters/src/utils/intersectExists.ts
```diff
@@ -0,0 +1,12 @@
...
+
+/**
+ * Check if two ranges have intersection
+ *
+ * @param firstRange - first text range
+ * @param secondRange - second text range
+ * @returns {boolean} true if itersect exists, false otherwise
+ */
+export function intersectionExists(firstRange: TextRange, secondRange: TextRange): boolean {
+  return (firstRange[0] < secondRange[1] && firstRange[0] > secondRange[0] || firstRange[1] < secondRange[1] && firstRange[1] > secondRange[0]);
```
gohabereg, line 11: You can use a formula
```
(firstRange[0] - secondRange[1]) * (secondRange[0] - firstRange[1]) > 0 
```

packages/dom-adapters/src/utils/intersectExists.ts
```diff
@@ -0,0 +1,12 @@
+import type { TextRange } from '@editorjs/model';
+
+/**
+ * Check if two ranges have intersection
+ *
+ * @param firstRange - first text range
+ * @param secondRange - second text range
+ * @returns {boolean} true if itersect exists, false otherwise
+ */
+export function intersectionExists(firstRange: TextRange, secondRange: TextRange): boolean {
```
gohabereg, line 10: Also let's rename to smth like `doRangesIntersect`

packages/dom-adapters/src/utils/mergeTextRanges.ts
```diff
@@ -0,0 +1,13 @@
...
+
+/**
+ * Return one merged from text range from two ranges
+ * This method does not check for intersection between ranges
+ *
+ * @param firstRange - first range to be merged
+ * @param secondRange - second range to be merged
+ * @returns {TextRange} merged text range
+ */
+export function mergeTextRanges(firstRange: TextRange, secondRange: TextRange): TextRange {
```
gohabereg, line 11: Would be great to check if ranges intersects right inside this function

packages/playground/src/components/Toolbar/index.ts
```diff
@@ -0,0 +1,120 @@
...
+       * Get current input with selection
+       */
+      if (selection) {
+        /**
+         * Do not render inline toolbar for not contenteditable elements
+         */
+        if (selection.focusNode?.nodeType !== Node.TEXT_NODE) {
+          return;
+        }
+      }
```
gohabereg, line 76: Would be enough to check if caret has textRange. If it has it — that means caret is in the text input

packages/dom-adapters/src/InlineToolAdapter/index.ts
```diff
@@ -0,0 +1,144 @@
...
+  intersectType: IntersectType,
+  create(data?: InlineToolData): HTMLElement;
+  getAction(index: TextRange, fragments: InlineFragment[], intersectType: IntersectType, data?: InlineToolData): { action: FormattingAction; range: TextRange };
+}
+
+/**
+ * Class handles on format model events and renders inline tools
+ * Applies format to the model
+ */
+export class InlineToolAdapter {
```
gohabereg, line 28: If it would be a single adapter for all of the tools, let's name it `InlineToolsAdapter`

packages/playground/src/components/Toolbar/index.ts
e11sy: as i can see, caret also has text range for native inputs, but i want to check that current node is text and contenteditable

packages/dom-adapters/src/InlineToolsAdapter/index.ts
```diff
@@ -0,0 +1,151 @@
...
+} from '@editorjs/model';
+import {
+  EventType,
+  TextFormattedEvent,
+  TextUnformattedEvent
+} from '@editorjs/model';
+import type { CaretAdapter } from '../CaretAdapter/index.js';
+import { type IntersectType, FormattingAction } from '@editorjs/model';
+
+export interface InlineTool {
```
neSpecc, line 17: use InlineTool from @editorjs/editorjs

packages/dom-adapters/src/InlineToolsAdapter/index.ts
```diff
@@ -0,0 +1,151 @@
...
+    this.#model.addEventListener(EventType.Changed, (event: ModelEvents) => this.#handleModelUpdates(event));
+  }
+
+  /**
+   * Handles text format and unformat model events
+   *
+   * @param event - model change event
+   */
+  #handleModelUpdates(event: ModelEvents): void {
+    if (event instanceof TextFormattedEvent || event instanceof TextUnformattedEvent) {
```
neSpecc, line 65:
```suggestion
    if (event instanceof TextFormattedEvent) {
```

packages/dom-adapters/src/InlineToolsAdapter/index.ts
```diff
@@ -0,0 +1,151 @@
...
+        const range = selection.getRangeAt(0);
+
+        const inlineElement = tool.create();
+
+        /**
+         * Insert contents from range to new inline element and put created element in range
+         */
+        inlineElement.appendChild(range.extractContents());
+        range.insertNode(inlineElement);
+      }
```
neSpecc, line 87: lets move it to `handleTextFormatEvent` method

packages/dom-adapters/src/InlineToolsAdapter/index.ts
```diff
@@ -0,0 +1,151 @@
...
+   * Detaches InlineTool from the adapter
+   *
+   * @param tool - tool to detach
+   */
+  public detachTool(tool: InlineTool): void {
+    this.#tools.delete(tool.name);
+  }
+
+  /**
+   * Applies format of the tool to the model
```
neSpecc, line 110: bad description

packages/dom-adapters/src/InlineToolsAdapter/index.ts
```diff
@@ -0,0 +1,151 @@
...
+    if (index === null) {
+      throw new Error('InlineToolsAdapter: caret index is outside of the input');
+    }
+
+    const textRange = index.textRange;
+    const blockIndex = index.blockIndex;
+    const dataKey = index.dataKey;
+
+    if (textRange === undefined || blockIndex === undefined || dataKey === undefined) {
+      return;
```
neSpecc, line 127: is it an exception?

packages/dom-adapters/src/utils/doRangesIntersect.ts
```diff
@@ -0,0 +1,12 @@
+import type { TextRange } from '@editorjs/model';
+
+/**
+ * Check if two ranges have intersection
+ *
+ * @param firstRange - first text range
+ * @param secondRange - second text range
+ * @returns {boolean} true if itersect exists, false otherwise
+ */
+export function doRangesIntersect(firstRange: TextRange, secondRange: TextRange): boolean {
```
neSpecc, line 10: add tests for this util

packages/dom-adapters/src/utils/mergeTextRanges.ts
```diff
@@ -0,0 +1,17 @@
...
+import { doRangesIntersect } from './doRangesIntersect.js';
+
+/**
+ * Merge two text ranges
+ *
+ * @param firstRange - first range to be merged
+ * @param secondRange - second range to be merged
+ * @returns {TextRange | null} merged text range or null if ranges do not intersect
+ */
+export function mergeTextRanges(firstRange: TextRange, secondRange: TextRange): TextRange | null {
```
neSpecc, line 11: also, tests are needed

packages/model/src/entities/inline-fragments/FormattingInlineNode/types/IntersectType.ts
```diff
@@ -0,0 +1,8 @@
+/**
+ * Types of interaction between two inline fragments with intersect ranges
+ */
+export enum IntersectType {
+  Extend = 'extend',
+  Replace = 'replace',
+  Separate = 'separate'
```
neSpecc, line 7: add docs

packages/playground/src/components/Toolbar/index.ts
```diff
@@ -0,0 +1,120 @@
+import type { InlineTool, InlineToolsAdapter } from '@editorjs/dom-adapters';
+import { type EditorJSModel, type TextRange, Index } from '@editorjs/model';
+import { EventType } from '@editorjs/model';
+import type { Nominal } from '@editorjs/model/dist/utils/Nominal';
+import { ref } from 'vue';
```
neSpecc, line 5: toolbar should not depend on vue

packages/dom-adapters/src/utils/doRangesIntersect.ts
neSpecc: we don't have tests for Dom-adapters so we can leave it for the future

packages/dom-adapters/src/InlineToolsAdapter/index.ts
e11sy: next pr probably

packages/core/src/entities/InlineTool.ts
```diff
@@ -0,0 +1,60 @@
+import type { InlineTool as InlineToolVersion2 } from '@editorjs/editorjs';
+import type { BlockToolConstructorOptions as BlockToolConstructorOptionsVersion2 } from '@editorjs/editorjs';
```
neSpecc, line 2: seems like unused imports

packages/core/src/entities/InlineTool.ts
```diff
@@ -0,0 +1,60 @@
...
+import type { TextRange, InlineFragment, FormattingAction, IntersectType, InlineToolName } from '@editorjs/model';
+
+/**
+ * Extended BlockToolConstructorOptions interface for version 3.
```
neSpecc, line 6: update docs

packages/core/src/entities/InlineTool.ts
```diff
@@ -0,0 +1,60 @@
...
+/**
+ * Block Tool interface for version 3
+ *
+ * In version 3, the save method is removed since all data is stored in the model
+ */
+export interface InlineTool extends Omit<InlineToolVersion2, 'save' | 'checkState' | 'render'> {
+  /**
+   * Name of the inline tool
+   */
+  name: InlineToolName;
```
neSpecc, line 20: usually we don't include name at the tools. Name is user-specific. It can specify it via the "tools" config property

packages/core/src/entities/InlineTool.ts
```diff
@@ -0,0 +1,60 @@
...
+
+  /**
+   * Returns formatting action and range for it to be applied
+   * @param index - index of current text selection
+   * @param fragments - all fragments of the bold inline tool inside of the current input
+   */
+  checkState(index: TextRange, fragments: InlineFragment[]): boolean;
+
+  /**
+   * Returns state of the bold inline tool
```
neSpecc, line 35: please, clarity description

packages/core/src/entities/InlineTool.ts
```diff
@@ -0,0 +1,60 @@
...
+    /**
+     * Range to apply formatting action
+     */
+    range: TextRange;
+  };
+
+  /**
+   * Method for rendering the tool
+   */
+  create(): HTMLElement;
```
neSpecc, line 54: what is the reason of renaming "render" to "create"?

packages/core/src/index.ts
```diff
@@ -41,6 +42,16 @@ export default class Core {
    */
   #caretAdapter: CaretAdapter;
 
+  /**
+   * Inline tool adapter is responsible for handling model updates
```
neSpecc, line 46: bad doc

packages/core/src/index.ts
```diff
@@ -41,6 +42,16 @@ export default class Core {
...
+   */
+  #inlineToolsAdapter: InlineToolsAdapter;
+
+  /**
+   * Inline toolbar is responsible for handling selection changes
```
neSpecc, line 51: clarify "handling" meaning

packages/core/src/tools/ToolsManager.ts
```diff
@@ -1,15 +1,22 @@
+import type { InlineTool } from '../entities/InlineTool.js';
 import type { BlockToolConstructor } from '../entities/BlockTool.js';
-import { Paragraph } from './internal/paragraph/index.js';
+import { Paragraph } from './internal/block-tools/paragraph/index.js';
 import type { EditorConfig } from '@editorjs/editorjs';
+import BoldInlineTool from './internal/inline-tools/Bold/index.js';
+import ItalicInlineTool from './internal/inline-tools/Italic/index.js';
```
neSpecc, line 6: lets make folder names consistent. "paragraph" does not have capitalizing

packages/core/src/entities/InlineTool.ts
```diff
@@ -0,0 +1,60 @@
...
+export interface InlineTool extends Omit<InlineToolVersion2, 'save' | 'checkState' | 'render'> {
+  /**
+   * Name of the inline tool
+   */
+  name: InlineToolName;
+
+  /**
+   * Type of merging of two ranges which intersect
+   */
+  intersectType: IntersectType;
```
neSpecc, line 25: it should be optional and have a default value

packages/core/src/entities/InlineTool.ts
```diff
@@ -0,0 +1,60 @@
...
+
+  /**
+   * Returns formatting action and range for it to be applied
```
neSpecc, line 28: doc is irrelevant

packages/core/src/tools/internal/inline-tools/Bold/index.ts
```diff
@@ -0,0 +1,80 @@
...
+    };
+  };
+
+  /**
+   * Returns state of the bold inline tool
+   * @param index - index of current selection
+   * @param fragments - all fragments of the bold inline tool inside of the current input
+   * @returns true if tool is active, false otherwise
+   */
+  public checkState(index: TextRange, fragments: InlineFragment[]): boolean {
```
neSpecc, line 61: should it be public?

neSpecc: also, it would be better to name method returning boolean starting with `is...` — `isActive` for example

packages/core/src/tools/internal/inline-tools/Italic/index.ts
neSpecc, line 1: same comments as in Bold

packages/core/src/tools/internal/inline-tools/Italic/index.ts
```diff
@@ -0,0 +1,81 @@
...
+  public checkState(index: TextRange, fragments: InlineFragment[]): boolean {
+    let isActive = false;
+
+    fragments.forEach((fragment) => {
+      /**
+       * Check if current index is inside of model fragment
+       */
+      if (index[0] >= fragment.range[0] && index[1] <= fragment.range[1]) {
+        isActive = true;
+        console.log(isActive, index, fragment);
```
neSpecc, line 70:
```suggestion
```

packages/core/src/tools/internal/inline-tools/Italic/index.ts
```diff
@@ -0,0 +1,81 @@
...
+
+        /**
+         * No need to check other fragments if state already chaned
```
neSpecc, line 73: need or don't need?

packages/core/src/ui/InlineToolbar/index.ts
```diff
@@ -0,0 +1,165 @@
...
+   * Toolbar html element related to the editor
+   */
+  #toolbar: HTMLElement | undefined = undefined;
+
+  /**
+   * @param model - editor model instance
+   * @param inlineToolAdapter - inline tool adapter instance
+   * @param tools - tools, that should be attached to adapter
+   */
+  constructor(model: EditorJSModel, inlineToolAdapter: InlineToolsAdapter, tools: InlineTool[]) {
```
neSpecc, line 46: ui should not depend on model. Its better to pass core EventBus that will send necessary events

packages/core/src/ui/InlineToolbar/index.ts
```diff
@@ -0,0 +1,165 @@
...
+  }
+
+  /**
+   * Handle changes of the caret selection
+   */
+  #handleSelectionChange(): void {
+    /**
+     * Listen to selection change ivents in model
+     */
+    this.#model.addEventListener(EventType.CaretManagerUpdated, (event) => {
```
neSpecc, line 63: We need to think about API between core and ui. For example, core will subscribe on model and emit SelectionChanged to the EventBus. And UI will depend only on EventBus

packages/core/src/ui/InlineToolbar/index.ts
```diff
@@ -0,0 +1,165 @@
...
+      const selection = window.getSelection();
+
+      /**
+       * Get current input with selection
+       */
+      if (selection) {
+        /**
+         * Do not render inline toolbar for not contenteditable elements
+         */
+        if (selection.focusNode?.nodeType !== Node.TEXT_NODE) {
```
neSpecc, line 73: are you sure that this statement is really for non-contenteditable elements?

packages/core/src/ui/InlineToolbar/index.ts
```diff
@@ -0,0 +1,165 @@
...
+        this.#toolbar.appendChild(inlineElementButton);
+      }
+    });
+
+    /**
+     * Get current input with selection
+     */
+    if (selection) {
+      /**
+       * Do not render inline toolbar for not contenteditable elements
```
neSpecc, line 140: are you sure that this statement is really for non-contenteditable elements?

packages/core/src/ui/InlineToolbar/index.ts
```diff
@@ -0,0 +1,165 @@
...
+       */
+      if (selection.focusNode !== null && selection.anchorNode !== null) {
+        selection.focusNode.parentElement?.parentNode?.insertBefore(this.#toolbar, selection.focusNode.parentElement);
```
neSpecc, line 143: >selection.focusNode.parentElement?.parentNode?

seems like unsafe call. What do you want to achieve?

packages/dom-adapters/src/InlineToolsAdapter/index.ts
```diff
@@ -0,0 +1,152 @@
...
+    this.#model.addEventListener(EventType.Changed, (event: ModelEvents) => this.#handleModelUpdates(event));
+  }
+
+  /**
+   * Handles text format and unformat model events
+   *
+   * @param event - model change event
+   */
+  #handleModelUpdates(event: ModelEvents): void {
+    console.log('model', event);
```
neSpecc, line 56:
```suggestion
```

packages/dom-adapters/src/InlineToolsAdapter/index.ts
```diff
@@ -0,0 +1,152 @@
...
+
+    if (!tool) {
+      throw new Error(`InlineToolsAdapter: tool ${toolName} is not attached`);
+    }
+
+    const fragments = this.#model.getFragments(blockIndex, dataKey, ...textRange, toolName);
+
+    const { action, range } = tool.getAction(textRange, fragments);
+
+    console.log('tool get action in adapter');
```
neSpecc, line 139: remove all logs please

packages/model/src/entities/inline-fragments/FormattingInlineNode/types/FormattingAction.ts
```diff
@@ -0,0 +1,4 @@
+export enum FormattingAction {
```
neSpecc, line 1: docs missed

packages/core/src/entities/InlineTool.ts
```diff
@@ -0,0 +1,60 @@
...
+  getAction(index: TextRange, fragments: InlineFragment[]): {
+    /**
+     * Formatting action - format or unformat
+     */
+    action: FormattingAction;
+
+    /**
+     * Range to apply formatting action
+     */
+    range: TextRange;
```
gohabereg, line 48: Let's extract this to a separate intereface

packages/core/src/ui/InlineToolbar/index.ts
gohabereg, line 73: I left same comment previously, you should be able to just check if index has a text range, no need to check node type

packages/core/src/ui/InlineToolbar/index.ts
neSpecc: we can do it in future requests

packages/core/src/ui/InlineToolbar/index.ts
neSpecc: we can do it in future requests

packages/model/src/entities/inline-fragments/FormattingInlineNode/types/IntersectType.ts
```diff
@@ -0,0 +1,19 @@
...
+
+  /**
+   * If two new fragment intersect with existing - remove existing with new one
+   */
+  Replace = 'replace',
+
+  /**
+   * If two fragments of one tool intersect - treat them as two different fragments
+   */
+  Separate = 'separate'
```
gohabereg, line 18: I don't really like that name, I think it's not descriptive enough. Maybe something like `LeaveBoth` or `None` as there's no intersection in that case. Happy to discuss other ideas

packages/core/src/entities/InlineTool.ts
e11sy: render() was rendering `MenuConfig` button, create method just creates tag for the tool

packages/core/src/tools/internal/inline-tools/Bold/index.ts
e11sy: it should be public because toolbar would need to know state of each tool

packages/core/src/ui/InlineToolbar/index.ts
e11sy: that state checks, that selection is in `TEXT_NODE` element, it means, that only selection in text contenteditable would render toolbar (native inputs do not support inline tags, but they also would have selection, this is why we need this condition)

packages/core/src/ui/InlineToolbar/index.ts
e11sy: this is temporary solution, i want to insert toolbar element before text element with selection, that could be changed, when we will think about complete inline toolbar

packages/core/src/ui/InlineToolbar/index.ts
neSpecc: put this description to the doc please

packages/core/src/entities/InlineTool.ts
e11sy: since we don't have ability to form inline tools collection from editor config, i need to pass tool names manually for correct attaching tools to adapter

packages/core/src/entities/InlineTool.ts
e11sy: is this related to config also?

packages/core/src/ui/InlineToolbar/index.ts
e11sy: updated comments

packages/model/src/entities/inline-fragments/FormattingInlineNode/types/IntersectType.ts
```diff
@@ -0,0 +1,19 @@
...
+
+  /**
+   * If two new fragment intersect with existing - remove existing with new one
+   */
+  Replace = 'replace',
+
+  /**
+   * If two fragments of one tool intersect - treat them as two different fragments
+   */
+  LeaveBoth = 'LeaveBoth'
```
gohabereg, line 18: To be consistent
```suggestion
  LeaveBoth = 'leave-both'
```

gohabereg (approved):

packages/core/src/entities/InlineTool.ts
```diff
@@ -0,0 +1,74 @@
+import type { InlineTool as InlineToolVersion2 } from '@editorjs/editorjs';
+import type { TextRange, InlineFragment, FormattingAction, IntersectType, InlineToolName } from '@editorjs/model';
+
+/**
+ * Extended InlineToolConstructorOptions interface for version 3.
+ */
+// eslint-disable-next-line @typescript-eslint/no-empty-object-type
+export interface InlineToolConstructorOptions extends InlineToolVersion2 {}
```
neSpecc, line 8: why constructor option extends tool?

packages/core/src/entities/InlineTool.ts
```diff
@@ -0,0 +1,74 @@
...
+  /**
+   * Method for creating wrapper element of the tool
+   */
+  createWrapper(): HTMLElement;
+}
+
+/**
+ * Interface, that represents inline tool with configured name
+ */
+export interface InlineToolWithName {
```
neSpecc, line 59:
```suggestion
export type InlineToolsConfig extends Record<string, InlineTool>
```

packages/core/src/tools/ToolsManager.ts
```diff
@@ -24,4 +31,17 @@ export default class ToolsManager {
...
+   * Returns inline tools got from the EditorConfig tools
+   */
+  public getInlineTools(): InlineToolWithName[] {
+    return [{
+      name: createInlineToolName('Bold'),
+      tool: new BoldInlineTool(),
+    }, {
+      name: createInlineToolName('Italic'),
+      tool: new ItalicInlineTool(),
+    }];
```
neSpecc, line 45:
```suggestion
    return {
      bold: BoldInlineTool,
      italic: ItalicInlineTool,
    };
```

class construction should be somewhere else

packages/core/src/tools/internal/inline-tools/bold/index.ts
```diff
@@ -0,0 +1,78 @@
...
+   * Specifies Tool as Inline Toolbar Tool
+   * @returns {boolean}
+   */
+  public static isInline = true;
+
+  /**
+   * @todo - get rid of tool name, since name could be user-spesific
+   * It is a temporary solution, because tools inline tools are now formed not from config
+   */
+  public name: InlineToolName = createInlineToolName('Bold');
```
neSpecc, line 25: redunand

packages/core/src/tools/internal/inline-tools/italic/index.ts
```diff
@@ -0,0 +1,78 @@
...
+   * Specifies Tool as Inline Toolbar Tool
+   * @returns {boolean}
+   */
+  public static isInline = true;
+
+  /**
+   * @todo - get rid of tool name, since name could be user-spesific
+   * It is a temporary solution, because tools inline tools are now formed not from config
+   */
+  public name: InlineToolName = createInlineToolName('Iatlic');
```
neSpecc, line 25: same

packages/playground/tsconfig.json
```diff
@@ -24,7 +24,7 @@
     },
     "baseUrl": ".",
   },
-  "include": ["src/**/*.ts", "src/**/*.d.ts", "src/**/*.tsx", "src/**/*.vue"],
+  "include": ["src/**/*.ts", "src/**/*.d.ts", "src/**/*.tsx", "src/**/*.vue", "src/components/Toolbar/Toolbarvue"],
```
neSpecc, line 27: seems irrelevant

packages/dom-adapters/src/InlineToolsAdapter/index.ts
```diff
@@ -0,0 +1,149 @@
...
+  InlineToolName,
+  ModelEvents
+} from '@editorjs/model';
+import {
+  EventType,
+  TextFormattedEvent
+} from '@editorjs/model';
+import type { CaretAdapter } from '../CaretAdapter/index.js';
+import { FormattingAction } from '@editorjs/model';
+import type { InlineTool } from '@editorjs/core/src/entities/InlineTool.js';
```
neSpecc, line 13: bad import

neSpecc (approved):

PR #75, opened 2024-08-28
feat(core): core module added

neSpecc:
Core package added

- initializes Model
- subscribes to model updates, watches BlockAdded event
- creates Adapters for Tool
- creates a Tool passing adapters in there
- adds a Block according to the model updates

Now we should pass blocks to the model via `initializeDocument()` method instead of constructor. This is needed to allow to subscribe on model update first


packages/core/src/entities/BlockTool.ts
```diff
@@ -0,0 +1,27 @@
...
+import type { BlockToolConstructorOptions as BlockToolConstructorOptionsVersion2 } from '@editorjs/editorjs';
+
+/**
+ * Extended BlockToolConstructorOptions interface for version 3.
+ */
+export interface BlockToolConstructorOptions extends BlockToolConstructorOptionsVersion2 {
+  /**
+   * Block tool adapter will be passed to the tool to connect data with the DOM
+   */
+  blockToolAdapter: BlockToolAdapter;
```
gohabereg, line 12: Maybe just `adapter` as this is a block tool by declaration

packages/core/src/index.ts
```diff
@@ -0,0 +1,144 @@
...
+    this.#model.initializeDocument({ blocks });
+  }
+
+  /**
+   * Validate configuration
+   * @param config - Editor configuration
+   */
+  private validateConfig(config: CoreConfig): void {
+    if (config.holder === undefined) {
+      const holder = document.getElementById('editorjs');
```
gohabereg, line 56: Magic string

packages/core/src/index.ts
```diff
@@ -0,0 +1,144 @@
...
+   * @param event - Block added event
+   */
+  private async handleBlockAdded(event: BlockAddedEvent): Promise<void> {
+    /**
+     * @todo add batch rendering to improve performance on large documents
+     */
+    const index = event.detail.index;
+
+    if (index.blockIndex === undefined) {
+      throw new Error('Block index should be defined');
```
gohabereg, line 99: If blockIndex is not there that means something is wrong with the model. This error would give a user no information on how to fix that

packages/core/src/index.ts
```diff
@@ -0,0 +1,144 @@
...
+    }
+
+    const caretAdapter = new CaretAdapter(this.#config.holder, this.#model);
```
gohabereg, line 102: CaretAdapter should be a single instance for the whole document, not for a block

packages/core/src/tools/ToolsManager.ts
```diff
@@ -0,0 +1,27 @@
+import type { BlockToolConstructor } from '../entities/BlockTool.js';
+import { Paragraph } from './internal/paragraph/index.js';
+import type { EditorConfig } from '@editorjs/editorjs';
+
+/**
+ * Works with tools
+ */
+export default class ToolsManager {
```
gohabereg, line 8: Is this something we intend to keep? We could use ToolsFactory/ToolsCollection from the 2nd version

packages/core/src/index.ts
```diff
@@ -0,0 +1,144 @@
...
+
+    const block = this.createBlock({
+      name: event.detail.data.name,
+      data: event.detail.data.data,
+    }, blockToolAdapter);
+
+    const blockEl = await block.render();
+
+    this.#config.holder.appendChild(blockEl);
+  }
```
e11sy, line 113: should new `blockEl` be inserted respectfully to model index?

packages/core/src/utils/composeDataFromVersion2.ts
```diff
@@ -0,0 +1,33 @@
...
+    blocks: data.blocks.map((block) => {
+      return {
+        name: block.type,
+        data: Object.fromEntries(
+          Object
+            .entries(block.data as Record<string, unknown>)
+            .map(([key, value]) => {
+              return [
+                key, {
+                  value: value as TextNodeSerialized,
```
gohabereg, line 24: Should be checked if typeof value is a string

packages/core/src/tools/internal/paragraph/index.ts
```diff
@@ -0,0 +1,32 @@
...
+ * Base text block tool
+ */
+export class Paragraph implements BlockTool {
+  /**
+   * Adapter for linking block data with the DOM
+   */
+  #adapter: BlockToolAdapter;
+  /**
+   * @param options - Block tool constructor options
+   */
```
e11sy, line 14:
```suggestion
  /**
   * Adapter for linking block data with the DOM
   */
  #adapter: BlockToolAdapter;
  
  /**
   * @param options - Block tool constructor options
   */
```

e11sy: also i think that it would be better to specify: #blockToolAdapter

packages/playground/src/App.vue
```diff
@@ -2,38 +2,46 @@
...
+          value: 'This is textarea element',
+          $t: 't',
+        },
       },
     },
-  } ],
+  ],
 });
-const document = ref(new EditorDocument(model.serialized));
+const eDocument = ref(new EditorDocument());
```
e11sy, line 41: maybe `editorDocument` ?

packages/core/src/entities/BlockTool.ts
neSpecc: done

packages/core/src/index.ts
neSpecc: I've left a todo for supporting inserting a block in middle

e11sy (approved):

gohabereg (approved):

PR #78, opened 2024-08-28
[BlockToolAdapter] Improve native inputs handling

gohabereg:


packages/dom-adapters/src/utils/findHardLineBoundary.ts
```diff
@@ -0,0 +1,39 @@
+/**
+ * Finds nearest next caret symbol from passed position
```
neSpecc, line 2:
```suggestion
 * Finds nearest next carriage return symbol from passed position
```

packages/dom-adapters/src/BlockToolAdapter/index.ts
```diff
@@ -181,7 +241,13 @@ export class BlockToolAdapter {
...
          * @see https://www.w3.org/TR/input-events-2/#overview
          */
-        const data = event.dataTransfer!.getData('text/plain');
+        let data: string;
+
+        if (isInputNative) {
+          data = event.data ?? '';
+        } else {
+          data = event.dataTransfer!.getData('text/plain');
+        }
```
e11sy, line 250: add coments for this cases please

e11sy (approved):

neSpecc (approved):

PR #85, opened 2024-08-29
Add BlockManager and EditorUI classes

gohabereg:


packages/core/src/ui/InlineToolbar/Editor/index.ts
e11sy, line 1: something wrong with paths, i guess it should be
UI 
- InlinteToolbar
- Editor

packages/core/src/BlockManager.ts
```diff
@@ -0,0 +1,76 @@
...
+    }
+
+    const block = tool.create({
+      adapter: blockToolAdapter,
+      data: data,
+      block: {} as BlockAPI,
+      readOnly: false,
+    });
+
+    this.#editorUI.addBlock(await block.render(), index.blockIndex);
```
neSpecc, line 64: add try-catch please

neSpecc (approved):

e11sy (approved):

PR #89, opened 2024-08-30
Add Core event bus

gohabereg:
EventBas for communication between modules


packages/core/src/components/BlockManager.ts
```diff
@@ -177,7 +177,12 @@ export class BlocksManager {
     try {
       const blockElement = await block.render();
 
-      this.#editorUI.addBlock(blockElement, index.blockIndex);
+      this.#eventBus.dispatchEvent(new BlockAddedCoreEvent({
+        tool: tool.name,
+        data: data.data,
+        uiContent: blockElement,
```
neSpecc, line 183: maybe `element`?

neSpecc (approved):

packages/core/src/components/BlockManager.ts
gohabereg: It could be not only HTMLElement, basically anything the tool returns

e11sy (approved):

PR #90, opened 2024-08-30
feat(playground): playground uses editor, input rendering

neSpecc:
1. Playground now visualizes Editor

image

2. BlockToolAdapter's `attachInput` now renders input initial content with formatting

image

3. Model now has `getText(dataKey)` method, Used for initial input value fill

4. Core's `prepareTools()` now awaited before initialization

5. FormattingAdapter `formatElementContent` fixed, previously it was throwing IndexSizeError

6. v2-converter now strips tags for `value`

### DIFF
--- a/packages/dom-adapters/src/FormattingAdapter/index.ts
+++ b/packages/dom-adapters/src/FormattingAdapter/index.ts
@@ -107,14 +107,8 @@ export class FormattingAdapter {
        */
       const range = document.createRange();
 
-      const inputTextNode = input.firstChild;
-
-      if (inputTextNode === null) {
-        throw new Error('FormattingAdapter: input element should contain text node');
-      }
-
-      range.setStart(inputTextNode, start);
-      range.setEnd(inputTextNode, end);
+      range.setStart(input, start);
+      range.setEnd(input, end);
 
       const inlineElement = tool.createWrapper(toolData);
 
--- a/packages/dom-adapters/src/FormattingAdapter/index.ts
+++ b/packages/dom-adapters/src/FormattingAdapter/index.ts
@@ -101,17 +101,27 @@ export class FormattingAdapter {
 
     const [start, end] = index;
 
-    /**
-     * Create range with positions specified in index
-     */
-    const range = document.createRange();
+    try {
+      /**
+       * Create range with positions specified in index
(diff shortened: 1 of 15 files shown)


gohabereg (approved):

packages/dom-adapters/src/FormattingAdapter/index.ts
```diff
@@ -101,17 +101,27 @@ export class FormattingAdapter {
...
+      const range = document.createRange();
 
-    range.setStart(input, start);
-    range.setEnd(input, end);
+      const inputTextNode = input.firstChild;
 
-    const inlineElement = tool.createWrapper(toolData);
+      if (inputTextNode === null) {
+        throw new Error('FormattingAdapter: input element should contain text node');
+      }
```
e11sy, line 114: pass checked text node to avoid this error catching

packages/model/src/entities/BlockNode/index.ts
```diff
@@ -191,6 +191,19 @@ export class BlockNode extends EventBus {
     node.update(value);
   }
 
+  /**
+   * Returns a text value of the specified text node
```
e11sy, line 195: Node is not specified for this function

packages/playground/src/App.vue
```diff
@@ -1,61 +1,24 @@
 <script setup lang="ts">
-import CaretIndex from '@/components/CaretIndex.vue';
-import { BlockToolAdapter, CaretAdapter, FormattingAdapter } from '@editorjs/dom-adapters';
-import { EditorDocument, EditorJSModel, EventType } from '@editorjs/model';
+// import CaretIndex from '@/components/CaretIndex.vue';
```
e11sy, line 2: seems redundant

packages/model/src/entities/BlockNode/index.ts
```diff
@@ -191,6 +191,19 @@ export class BlockNode extends EventBus {
     node.update(value);
   }
 
+  /**
+   * Returns a text value of the specified data key
```
e11sy, line 195:
```suggestion
   * Returns a text value of the input with specified data key
```

e11sy (approved):

PR #92, opened 2024-08-31
Implement transformations

nikmel2803:
- implement logic for operations transformations (text operations only for now)
- move logic from the `Transformer` class to the `Operation` class


packages/collaboration-manager/src/Operation.spec.ts
```diff
@@ -0,0 +1,164 @@
...
+    return new Operation(
+      type,
+      new IndexBuilder()
+        .addBlockIndex(0)
+        .addDataKey('text' as DataKey)
+        .addTextRange([startIndex, startIndex])
+        .build(),
+      {
+        prevValue: type === OperationType.Delete ? value : '',
+        newValue: type === OperationType.Insert ? value : '',
```
gohabereg, line 17: Pls remove this for Add/Remove opeeration, leave just `data`

packages/collaboration-manager/src/Operation.ts
```diff
@@ -56,4 +56,130 @@ export class Operation {
...
+    }
+
+    if (this.type === OperationType.Modify || againstOp.type === OperationType.Modify) {
+      throw new Error('Unsupported operation type');
+    }
+
+    /**
+     * Do not transform operations if they are on different blocks or documents
+     */
+    if (this.index.documentId !== againstOp.index.documentId || this.index.blockIndex !== againstOp.index.blockIndex) {
```
gohabereg, line 116: We need to consider block index for text operations, if block before was added or deleted, we need to update  the index

packages/collaboration-manager/src/Operation.ts
```diff
@@ -56,4 +56,130 @@ export class Operation {
...
+      return this;
+    }
+
+    const [ receivedStartIndex ] = this.index.textRange!;
+    const [ localStartIndex ] = againstOp.index.textRange!;
+
+    switch (true) {
+      case this.type === OperationType.Insert && againstOp.type === OperationType.Insert:
```
gohabereg, line 124: The type of the current operation doesn't matter here, index will be changed the same way for both insert and delte

packages/collaboration-manager/src/Operation.ts
gohabereg: Can be done in a separate PR

packages/collaboration-manager/src/Operation.ts
gohabereg: There are only three cases here:
1. If received operation is delete operatio **before** (index-wise) the current, we need to update the current one index shifting it left
2. If received operation is insert operation **before** (index-wise) the current, we need to shift index right
3. If operation of any type happened **after** (index-wise) the current, no need to do anything

packages/collaboration-manager/src/Operation.spec.ts
```diff
@@ -0,0 +1,164 @@
...
+
+    test('Should adjust an index for a received operation if it is after a local one', () => {
+      const receivedOp = createOperation(OperationType.Insert, 3, 'def');
+      const localOp = createOperation(OperationType.Insert, 0, 'abc');
+      const transformedOp = receivedOp.transform(localOp);
+
+      expect(transformedOp.index.textRange).toEqual([6, 6]);
+    });
+
+    test('Should not change a received operation if it is at the same position as a local one', () => {
```
gohabereg, line 39: If position is the same, the index should change by the length of text in received operation

packages/collaboration-manager/src/CollaborationManager.ts
```diff
@@ -111,17 +119,40 @@ export class CollaborationManager {
...
         operation = new Operation(OperationType.Delete, e.detail.index, {
-          prevValue: e.detail.data,
-          newValue: '',
+          payload: e.detail.data,
+        });
+        break;
+      case (e instanceof TextFormattedEvent):
+        operation = new Operation(OperationType.Modify, e.detail.index, {
+          payload: e.detail.data,
+          prevPayload: undefined,
```
neSpecc, line 139: maybe null?

packages/collaboration-manager/src/Operation.spec.ts
```diff
@@ -0,0 +1,318 @@
...
+
+    it('should throw an error if unsupported operation type is provided', () => {
+      const receivedOp = createOperation(OperationType.Insert, 0, 'def');
+      // @ts-expect-error — for test purposes
+      const localOp = createOperation('unsupported', 0, 'def');
+
+      expect(() => receivedOp.transform(localOp)).toThrow('Unsupported operation type');
+    });
+
+    it('should not transform relative the Modify operation', () => {
```
neSpecc, line 83: unclear test case name

packages/collaboration-manager/src/Operation.ts
```diff
@@ -9,30 +9,51 @@ export enum OperationType {
   Modify = 'modify'
 }
 
+type OperationPayload = string | BlockNodeSerialized;
```
neSpecc, line 12: jsdoc with simple examples will be helpful

packages/collaboration-manager/src/Operation.ts
```diff
@@ -9,30 +9,51 @@ export enum OperationType {
...
    */
-  prevValue: string;
+  payload?: T;
 
   /**
-   * Value after the operation
+   * Previous payload for undo/redo purposes
    */
-  newValue: string;
+  prevPayload?: T;
```
neSpecc, line 35: T | null ?

packages/collaboration-manager/src/UndoRedoManager.spec.ts
```diff
@@ -0,0 +1,101 @@
...
+    expect(invertedOp).toEqual(op.inverse());
+  });
+
+  it('should return undefined on undo if there is no operations in stack', () => {
+    const manager = new UndoRedoManager();
+
+    expect(manager.undo()).toBeUndefined();
+  });
+
+  it('should return undefined on undo if there is no operations in stack', () => {
```
e11sy, line 35:
```suggestion
  it('should return undefined on redo if there is no operations in stack', () => {
```

packages/collaboration-manager/src/Operation.ts
```diff
@@ -51,9 +72,119 @@ export class Operation {
...
+
+    return new Operation(
+      this.type,
+      newIndexBuilder.build(),
+      this.data
+    );
+  }
+
+  /**
+   * Checks if operation needs to be transformed
```
neSpecc, line 175: Describe the cases, when it should be transformed

e11sy (approved):

neSpecc (approved):

PR #95, opened 2025-04-06
imp(dom-adapters): rerender related nodes on inline formatting

e11sy:
## Changes
- now `dom-adapter` rerenders related nodes on model changes
- add utility function that expands node boundaries for the range


packages/dom-adapters/src/FormattingAdapter/index.ts
```diff
@@ -171,9 +173,72 @@ export class FormattingAdapter {
...
 
-      const inlineElement = tool.createWrapper(event.detail.data.data);
+      const affectedFragments = this.#model.getFragments(blockIndex, dataKey, ...textRange);
 
-      surround(inlineElement, input, textRange);
+      let lowerBoundary = textRange[0];
+      let upperBoundary = textRange[1];
+
+      for (const fragment of affectedFragments) {
+        lowerBoundary = Math.min(lowerBoundary, fragment.range[0] ?? 0);
```
gohabereg, line 182: Model returns fragments in order, so the left boundary will always be the first fragment

packages/dom-adapters/src/FormattingAdapter/index.ts
```diff
@@ -171,9 +173,77 @@ export class FormattingAdapter {
...
+      const inputContent = input.textContent;
 
-      surround(inlineElement, input, textRange);
+      const rangeStart = Math.max(0, textRange[0] - 1)
+      const rangeEnd = inputContent !== null ? Math.min(inputContent.length, textRange[1] + 1) : 0;
+
+      const affectedFragments = this.#model.getFragments(blockIndex, dataKey, rangeStart, rangeEnd);
+
+      const lowerBoundary = affectedFragments[0]?.range[0] ?? textRange[0];
+      let upperBoundary = textRange[1];
```
gohabereg, line 184: Let's do `leftBoundary` and `rightBoundary`

packages/dom-adapters/src/FormattingAdapter/index.ts
```diff
@@ -171,9 +173,77 @@ export class FormattingAdapter {
...
+    }
+
+    const extractedContent = range.cloneContents();
+
+    /**
+     * Create temporary container to allow formatting of the extracted content
+     */
+    const template = document.createElement('template');
+
+    template.innerHTML = extractedContent.textContent!;
```
gohabereg, line 226: Would be more efficient as no need to clone the DOM nodes

```suggestion
    /**
     * Create temporary container to allow formatting of the extracted content
     */
    const template = document.createElement('template');

    template.innerHTML = range.toString()!;
```

packages/dom-adapters/src/utils/expandRangeNodeBoundary.ts
```diff
@@ -0,0 +1,33 @@
...
+        return node;
+      }
+    }
+  } else {
+    while (!(node.parentNode instanceof HTMLElement && node.parentNode.getAttribute('contenteditable') === 'true')) {
+      if (node.parentNode && node === node.parentNode.firstChild) {
+        node = node.parentNode;
+      } else {
+        return node;
+      }
```
gohabereg, line 28: Could reduce nesting here as well

gohabereg (approved):

neSpecc (approved):

PR #104, opened 2025-04-10
chore(ui,core,sdk,dom-adapters): dom-adapters depends on event bus

neSpecc:
When there will be a parent contenteditable, we'll listed `beforeinput` on it, then dispatch `ui:before-input` event. BlockToolAdapter will handle it.

This PR contains preparations for that:
1) BlockToolAdapter now depends on EventBus
2) To prevent circular dependency (Dom-Adapters <-> Core), `EventBus` moved from Core to SDK
3) UI, SDK, Core, Dom Adapters now can be build and used as bundles instead of "src"
4) Various build fixes


packages/ui/src/Blocks/Blocks.ts
```diff
@@ -74,11 +75,11 @@ export class BlocksUI implements EditorjsPlugin {
 
       this.#eventBus.dispatchEvent(new Event('core:undo'));
     });
-
   }
 
   /**
    * Prepares blocks holder element
+   * @param editorHolder
```
gohabereg, line 82: Description is missing

gohabereg (approved):

e11sy (approved):

PR #105, opened 2025-04-10
imp(dom-adapters): save formatting on block split

e11sy:
## Changes
- Now fragments from extracted part of input would be moved to the new inserted block with adjusted text ranges
- Now CaretAdapter covers the case, when model selection and dom selection are in different inputs

### DIFF
--- a/packages/dom-adapters/src/BlockToolAdapter/index.ts
+++ b/packages/dom-adapters/src/BlockToolAdapter/index.ts
@@ -365,7 +365,7 @@ export class BlockToolAdapter implements BlockToolAdapterInterface {
      * Fragment ranges bounds should be decreased by end index, because end is the index of the first character of the new block
      */
     relatedFragments.forEach(fragment => {
-      fragment.range[0] = Math.max(0, fragment.range[0] - start);
+      fragment.range[0] = Math.max(0, fragment.range[0] - end);
       fragment.range[1] -= end;
     });
 


gohabereg (approved):

packages/dom-adapters/src/BlockToolAdapter/index.ts
```diff
@@ -351,6 +351,16 @@ export class BlockToolAdapter implements BlockToolAdapterInterface {
     const currentValue = this.#model.getText(this.#blockIndex, key);
     const newValueAfter = currentValue.slice(end);
 
+    const relatedFragments = this.#model.getFragments(this.#blockIndex, key, end, currentValue.length);
+
+    /**
+     * Adjust fragments ranges respectfully to the removed text
+     */
+    relatedFragments.forEach(fragment => {
+      fragment.range[0] -=end;
```
gohabereg, line 360:
```suggestion
      fragment.range[0] -= Math.max(0, end);
```

packages/dom-adapters/src/BlockToolAdapter/index.ts
```diff
@@ -351,6 +351,16 @@ export class BlockToolAdapter implements BlockToolAdapterInterface {
     const currentValue = this.#model.getText(this.#blockIndex, key);
     const newValueAfter = currentValue.slice(end);
 
+    const relatedFragments = this.#model.getFragments(this.#blockIndex, key, end, currentValue.length);
+
+    /**
+     * Adjust fragments ranges respectfully to the removed text
```
neSpecc, line 357: I can't understand what does it mean "adjust". Could you write more detailed explanation please?

neSpecc (approved):

PR #107, opened 2025-04-10
Implement basic memory management for OT server

gohabereg:
Until we implement some database for OT server we need to manage memory usage

- When a new client connects to the server, it passes it's document state
- If it's the first client for the document, OT server initalizes the document with passed state
- If it's no the first client, OT sends the document current state to the client

Also added Dockerfile and compose config for OT server


e11sy (approved):

compose.yaml
```diff
@@ -0,0 +1,20 @@
+# Comments are provided throughout this file to help you get started.
+# If you need more help, visit the Docker compose reference guide at
+# https://docs.docker.com/go/compose-spec-reference/
+
+# Here the instructions define your application as a service called "server".
+# This service is built from the Dockerfile in the current directory.
+# You can add other services your application may depend on here, such as a
+# database or a cache. For examples, see the Awesome Compose repository:
+# https://github.com/docker/awesome-compose
```
neSpecc, line 9: comments looks not so actual. Make them actual or remove them

packages/model/src/entities/EditorDocument/EditorDocument.spec.ts
```diff
@@ -50,6 +50,74 @@ describe('EditorDocument', () => {
...
+    });
+  });
+
+  describe('.clear()', () => {
+    it('should clear the document', () => {
+      const doc = createEditorDocumentWithSomeBlocks();
+
+      doc.clear();
+
+      console.log(doc.serialized);
```
neSpecc, line 115: rm doc

PR #108, opened 2025-04-10
chore(sdk): sdk now stores shared deps

neSpecc:
Second part of common `beforeinput` event.

The `sdk` package now stores all deps required by other packages


packages/sdk/package.json
```diff
@@ -1,15 +1,16 @@
 {
```
gohabereg, line 1: There's no CI workflow for sdk package, please add it

gohabereg (approved):

e11sy (approved):

PR #118, opened 2025-04-12
feat(ui, dom-adapters): cross-input selection

neSpecc:
1. UI blocks wrapper now contenteditable which enables cross-input selection (CIS)
2. BlockToolAdapter (all of them for each input) delegates `ui:before-input` to all affected inputs. To find them, it checks current range and compare its start/end with input position.
3. Delete handler updated: now it has 4 cases: 
  - when CIS starts in inout
  - when CIS ends in input
  - when input in-between CIS
  - when range is inside the input
4. Enter (Split) logic improved: it is being applied only for cases:
  - when CIS starts in inout
  - when range is inside the input
5. On any inputType we first remove selected range of CIS
6. CaretAdapter now stores `this.#blocks` instead of `this.#inputs`. Thus, we won't care about managing input indexes (with block index) on block removing/moving.


packages/dom-adapters/src/CaretAdapter/index.ts
```diff
@@ -88,22 +88,29 @@ export class CaretAdapter extends EventTarget {
...
    *
-   * @param index - index of the input to remove
+   * @param index - index of the block to remove
    */
-  public detachInput(index: Index): void {
-    this.#inputs.delete(index.serialize());
+  public detachBlock(index: Index): void {
+    const block = this.getBlock(index);
+
+    if (block) {
```
gohabereg: Reverse to reduce nesting

packages/dom-adapters/src/CaretAdapter/index.ts
```diff
@@ -130,30 +137,47 @@ export class CaretAdapter extends EventTarget {
...
+   * Finds block by index
    *
-   * @param index - index of the input in the model tree
+   * @param index - index of the block in the model tree
    */
-  public getInput(index?: Index): HTMLElement | undefined {
+  public getBlock(index?: Index): BlockToolAdapter | undefined {
+    if (index === undefined) {
+      if (this.#currentUserCaret.index === null) {
+        throw new Error('[CaretManager] No index provided and no user caret index found');
```
gohabereg, line 149: Maybe just return null?

e11sy (approved):

packages/dom-adapters/src/CaretAdapter/index.ts
neSpecc: not sure it is a normal case when getBlock us called wihtout index or user caret. Let's leave exception while we will be sure it is a normal case.

packages/dom-adapters/src/CaretAdapter/index.ts
neSpecc: done

gohabereg (approved):

PR #123, opened 2026-04-15
fix(formatting): apply formatting by shortcuts

neSpecc:
1. Fixed problem when selected fragment disappears when CMD+B (and other) applied.

2. Refactoring of Tools configuration:

`options` - set of feature flags and mappings for tool. Can be specified by tool internally and overrides by user on tool connection:

Internal predefined options:

```ts
class BoldInlineTool { 
  static get options() {
    return {
      config: { // example of tool internal configuration 
        defaultWeight: 500 
      },
      shortcut: 'CMD+B' // mappings options for Shortcuts plugin
    }
  }
}
```

Overrides 

```ts
core.use(BoldInlineTool, {
  config: {
    defaultWeight: 600,
  },
  shortcut: 'CMD+SHIFT+B'
})
```

3. Add ShortcutsPlugin that listens for `ui:key-down` event and fire correspondent plugin if matched.


packages/core/src/components/SelectionManager.ts
```diff
@@ -68,7 +73,7 @@ export class SelectionManager {
 
       this.#inlineTools.set(name, toolInstance);
 
-      this.#formattingAdapter.attachTool(name, toolInstance);
+      this.#formattingAdapter.attachTool(name, toolInstance, tool.shortcut);
```
gohabereg: Core shouldn't know anything about shortcuts — it's only UI/UX feature

gohabereg: FormattingAdapter should be the place as it's a DOM-adapter

packages/core/src/plugins/ShortcutsPlugin.ts
```diff
@@ -0,0 +1,135 @@
...
+        return;
+      }
+
+      for (const [shortcut, toolName] of this.#shortcutToToolName) {
+        if (matchKeyboardShortcut(nativeEvent, shortcut)) {
+          nativeEvent.preventDefault();
+
+          try {
+            this.#processInlineTool(toolName);
+          } catch {
```
gohabereg: You already have try/catch inside #processInlineTool method

packages/core/src/plugins/ShortcutsPlugin.ts
```diff
@@ -0,0 +1,135 @@
...
+
+  /**
+   * Registers `shortcut` from merged inline tool `options` (Editor.js codex string, e.g. CMD+B).
+   * @param toolName - name of the inline tool to apply
+   */
+  #processInlineTool(toolName: string): void {
+    try {
+      this.#api.selection.applyInlineToolForCurrentSelection(toolName as InlineToolName);
+    } catch {
+      /**
```
gohabereg: Maybe we can filter by the error and re-throw in case the error is not related to caret not being in the input?

Todo will be ok

packages/core/src/index.ts
```diff
@@ -164,11 +166,15 @@ export default class Core {
...
           id: pluginType,
           multiple: true,
-          value: [pluginOrTool, toolConfig],
+          value: [pluginOrTool, options],
         });
         break;
       default:
-        this.#iocContainer.set(pluginType, pluginOrTool);
+        this.#iocContainer.set({
+          id: PluginType.Plugin,
```
gohabereg, line 174: Why not leave `pluginType` there?

packages/core/src/index.ts
```diff
@@ -164,11 +166,15 @@ export default class Core {
...
+          multiple: true,
+          value: pluginOrTool,
```
gohabereg, line 176: Maybe support options for plugins as well? Also could be a todo

packages/sdk/src/entities/EntityType.ts
```diff
@@ -43,4 +43,14 @@ export enum ToolType {
...
+ * IoC keys and related ids for editor plugins (`core.use(MyPlugin)`).
+ */
+export enum PluginType {
+  /**
+   * Default plugin type
+   */
+  Plugin = 'Plugin'
+}
+
+export type EntityType = ToolType | UiComponentType | PluginType;
```
gohabereg, line 56: If we don't use UiComponentType inside the core, we can remove it. So all the plugins would just have `plugin` type

packages/core/src/index.ts
neSpecc: Explicit semantics: `pluginType` here means “what came from outside (from the constructor)”, while id: `PluginType.Plugin` is “the key under which we register it in the container”. Even if the constructor mistakenly sets a different type, the registration will still be done under the stable ID for plugins.

packages/core/src/plugins/ShortcutsPlugin.ts
neSpecc: fixed

packages/core/src/plugins/ShortcutsPlugin.ts
neSpecc: added

packages/core/src/index.ts
neSpecc: todo left

packages/sdk/src/entities/EntityType.ts
neSpecc: removed

neSpecc: returned back. It is used to connect `ToolboxUI`

```ts
 core
    .use(EditorjsUI)
    .use(BlocksUI)
    .use(InlineToolbarUI)
    .use(ToolboxUI)
    .initialize();
```

gohabereg: Yes, but now there's no reason to use it instead of just PluginType.Plugin

We can do it later though

gohabereg (approved):

e11sy (approved):

PR #124, opened 2026-04-16
feat(): attach and detach value

e11sy:
1. Allows to attach value via DOM-adapter
2. Operations transformer now can transform modify operations with data indexes
- if changes same block and same dataKey — checks revision, only latest modify should be applied (undefined revision is latest because it did not even send to the server yet)
3. `OTClient` now does not call collaborationManager callback (`applyOperation`) if new operation from the server transforms against `pendingOperations` into `Neutral`

so `CollaborationManager` waits only for valuable operations from `OTClient`

4. `detachValue` just removes model value node
5. added `detachValue`, `detachInput`, `attachValue` to the interface of the `BlockToolAdapter` in sdk


packages/collaboration-manager/src/client/OTClient.ts
```diff
@@ -226,6 +226,10 @@ export class OTClient {
 
     this.#rev = operation.rev!;
 
+    if (transformedOperation.type === OperationType.Neutral) {
```
gohabereg, line 222: Add a comment please why we don't apply a neutral op

packages/collaboration-manager/src/OperationsTransformer.ts
```diff
@@ -198,6 +201,47 @@ export class OperationsTransformer {
...
+      return Operation.from(operation);
+    }
+
+    // Cover case 2.1
+    if (againstOp.rev === undefined) {
+      return new Operation(OperationType.Neutral, index, { payload: [] }, operation.userId, operation.rev);
+    }
+
+    // Cover case 2.2
+    return Operation.from(operation);
```
gohabereg, line 245: We should check here if againstOp.rev >= operation.rev

packages/dom-adapters/src/BlockToolAdapter/index.ts
```diff
@@ -172,6 +178,92 @@ export class BlockToolAdapter implements BlockToolAdapterInterface {
...
+   */
+  public attachValue<T>(keyRaw: string, initialValue: T, callback: (value: T) => void): (newValue: T) => void {
+    const key = createDataKey(keyRaw);
+
+    /**
+     * Wrap callback to cast value to undefined and support generic type of the attachValue method
+     */
+    this.#attachedValues.set(key, (value) => {
+      callback(value as T);
+    });
```
gohabereg: Not clear why the wrap is needed. There's no cast to undefined

packages/collaboration-manager/src/client/OTClient.spec.ts
```diff
@@ -0,0 +1,244 @@
...
+describe('OTClient', () => {
+  describe('connect (mocked WebSocket)', () => {
+    const collabWsUrl = 'ws://test-collab.invalid/document';
+
+    type WsListener = (event: { data: string }) => void;
+
+    /**
+     * Minimal WebSocket double so the test can act as OT server (handshake + remote ops).
+     */
+    class MockWebSocket {
```
gohabereg: Maybe move the mock to a separate file

neSpecc (approved):

gohabereg (approved):

PR #126, opened 2026-04-16
feat(cis): composite index support

neSpecc:
## 1. Composite indexes support

Index for single input selection:

image

```
"block@0:data@text:[1,3]"
```

Index for cross-input selection (new):

image

```
{
  "composite": [
    "\"block@0:data@text:[1,3]\"",
    "\"block@1:data@text:[0,3]\"",
    "\"block@2:data@text:[0,2]\""
  ]
}
```

## 2. New util `selectionRangeInInput.ts`:
- checks for selection start/ends/contains inputs (moved from BlockToolAdapter)
- new method `getClippedTextRangeForInput` that returns `start` and `end` position of range inside an input  — logic moved from BlockToolAdapters `handleDeleteInContenteditable` (which is refactored a little)


packages/dom-adapters/src/CaretAdapter/index.ts
```diff
@@ -230,59 +268,75 @@ export class CaretAdapter extends EventTarget {
...
+      const blockB = b.blockIndex;
+
+      if (blockA !== blockB) {
+        return (blockA ?? 0) - (blockB ?? 0);
+      }
+
+      const keyA = a.dataKey !== undefined ? String(a.dataKey) : '';
+      const keyB = b.dataKey !== undefined ? String(b.dataKey) : '';
+
+      return keyA.localeCompare(keyB);
```
gohabereg: We shouldn't change order of the data nodes as for now we assume they are in order.  It seems tools would have to provide the order of the inputs

gohabereg (approved):

packages/dom-adapters/src/CaretAdapter/index.ts
neSpecc: changed for sorting by real Dom positions

e11sy (approved):

PR #132, opened 2026-05-01
Enable full keypath support for write operations

gohabereg:
All BlockNode operations now support keypath:

```
blockNode.insertText('compound.key.path', 'text') // serialized value gonna look like { compound: { key: { path: 'text' } }
```


packages/model/src/entities/BlockNode/BlockNode.spec.ts
```diff
@@ -532,6 +540,208 @@ describe('BlockNode', () => {
...
+      const blockNode = createBlockNodeWithData({});
+      const key = createDataKey('items.0');
+      const value = 'first item';
+
+      blockNode.createDataNode(key, value);
+
+      expect(get(blockNode.data, 'items.0')).toBeInstanceOf(ValueNode);
+    });
+
+    it('should create value node in a nested object inside an array', () => {
```
ilyamore88, line 557: Might there be array of text nodes? For example in a list tool. If it is possible case,  would be great to create test case for array of text nodes as well

packages/model/src/entities/BlockNode/BlockNode.spec.ts
```diff
@@ -532,6 +540,208 @@ describe('BlockNode', () => {
...
+      const blockNode = createBlockNodeWithData({});
+      const key = createDataKey('items.0.content');
+      const value = 'content text';
+
+      blockNode.createDataNode(key, value);
+
+      expect(get(blockNode.data, 'items.0.content')).toBeInstanceOf(ValueNode);
+    });
+
+    it('should not change the node if a nested key already exists', () => {
```
ilyamore88: Shouldn't it be the case with some thrown error?

packages/model/src/entities/BlockNode/BlockNode.spec.ts
```diff
@@ -532,6 +540,208 @@ describe('BlockNode', () => {
...
+        fragments: [],
+      });
+
+      const items = (blockNode.data as Record<string, unknown[]>)['items'];
+
+      expect(items).toHaveLength(2);
+      expect(items[1]).toStrictEqual(originalNode);
+    });
+
+    it('should always insert into an array even when an element already exists at that index', async () => {
```
ilyamore88: Same case as previous two. But you expect listener to be called. Remove or change description for test case.

packages/model/src/entities/BlockNode/BlockNode.spec.ts
```diff
@@ -532,6 +540,208 @@ describe('BlockNode', () => {
...
+
+      const items = (blockNode.data as Record<string, unknown[]>)['items'];
+
+      // new node was spliced in, old node shifted to index 1
+      expect(items).toHaveLength(2);
+      expect(listener).toHaveBeenCalled();
+    });
+  });
+
+  describe('.getDataNode()', () => {
```
ilyamore88, line 661: Already existed block. Move test cases to existed one.

packages/model/src/entities/BlockNode/BlockNode.spec.ts
```diff
@@ -532,6 +540,208 @@ describe('BlockNode', () => {
...
+    it('should return undefined if the key does not exist', () => {
```
ilyamore88, line 662: Maybe some test case for non-existed element of array as well?

packages/model/src/entities/BlockNode/BlockNode.spec.ts
ilyamore88: Covered by insertText() method test case later, but I still think it might be useful here as well.

packages/model/src/entities/BlockNode/BlockNode.spec.ts
```diff
@@ -1503,6 +1809,8 @@ describe('BlockNode', () => {
           tuneKey: key,
           tuneName: tuneName,
         }));
+      expect(event)
```
ilyamore88, line 1740: Looks like a mistake. No change in test case, but expected result has been changed.

ilyamore88 (changes requested):

packages/model/src/entities/BlockNode/BlockNode.spec.ts
gohabereg: Just an additional check for the user id

packages/model/src/entities/BlockNode/errors/AlreadyExistingKeyError.ts
```diff
@@ -0,0 +1,14 @@
+import type { DataKey } from '../types/index.js';
+
+/**
+ *
+ */
```
ilyamore88, line 5: Doc

packages/model/src/utils/keypath.spec.ts
```diff
@@ -1,10 +1,75 @@
...
+    it('should accept keys as an array', () => {
+      const object: Record<string, any> = { a: ['x', 'z'] };
+
+      insert(object, ['a', '1'], 'y');
+
+      expect(object.a).toEqual(['x', 'y', 'z']);
+    });
+  });
+
+  describe('remove()', () => {
```
ilyamore88: You created remove() block twice

ilyamore88 (approved):
Two minor requests. But I’m fine to approve it.

neSpecc (approved):

PR #141, opened 2026-05-09
Implement split block API

gohabereg:
Split moved to the core and enhanced. Accessible via Blocks API

Now split follows this algorithm:
1. If a Tool that is being split has `canSplit: true` option:
a. Core would split the inputs in creation order and insert a new block of the same type with inputs after the split. That means if a Tool provides canSplit = true, it should be able to handle partial data
b. If inputs are contained in the array, array indexes will be renumbered so new block receives array starting from 0

2. If a Tool doesn't have `canSplit` or it equals false
a. New block of the `defaultTool` type inserted 
b. Content from the inputs after the split is concatenated into a string. Fragments also concatenated into a one array with ranges shifted
c. Default Tool *must* support `import` option in the `conversionConfig`
d. Content is converted using the conversion config and inserted into the new block

### DIFF
--- a/packages/core/src/components/BlockManager.ts
+++ b/packages/core/src/components/BlockManager.ts
@@ -242,6 +242,12 @@ export class BlocksManager {
 
     const [, splitInput] = blockTextContent[splitIndex];
 
+    if (offset < 0 || offset > splitInput.value.length) {
+      throw new RangeError(
+        `Offset ${offset} is out of range for input "${dataKey}" with length ${splitInput.value.length}`
+      );
+    }
+
     /**
      * Text and fragments from the input that was split
      */
--- a/packages/core/src/components/BlockManager.ts
+++ b/packages/core/src/components/BlockManager.ts
@@ -1,17 +1,22 @@
 import {
+  BlockChildType,
+  type BlockId,
   type BlockNodeInit,
+  type DataKey,
   type EditorDocumentSerialized,
-  EditorJSModel
+  EditorJSModel,
+  type InlineTreeNodeSerialized,
+  keypath,
+  mergeTextNodes,
+  NODE_TYPE_HIDDEN_PROP,
+  sliceFragments
 } from '@editorjs/model';
 import 'reflect-metadata';
 import { inject, injectable } from 'inversify';
 import { TOKENS } from '../tokens.js';
 import ToolsManager from '../tools/ToolsManager.js';
 import { BlockToolData } from '@editorjs/editorjs';
-import {
-  CoreConfigValidated,
-  EventBus
-} from '@editorjs/sdk';
+import { CoreConfigValidated, EventBus } from '@editorjs/sdk';
 
 /**
  * Parameters for the BlocksManager.insert() method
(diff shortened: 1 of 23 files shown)


packages/core/src/components/BlockManager.ts
```diff
@@ -207,6 +212,150 @@ export class BlocksManager {
...
+         * @todo check if \n is a proper option here
+         */
+        value: textAfter + '\n',
+        fragments: fragmentsAfter,
+      };
+
+      const contentAfter = mergeTextNodes(entriesAfter, contentAfterAccInit);
+
+      const defaultTool = this.#toolsManager.blockTools.get(this.#config.defaultBlock)!;
+
```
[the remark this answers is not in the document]
gohabereg: There's a todo to implement validation on initialisation later

packages/sdk/src/tools/facades/BlockToolFacade.ts
```diff
@@ -153,6 +155,37 @@ export class BlockToolFacade extends BaseToolFacade<ToolType.Block, IBlockTool>
...
+      return importFnOrProp(value);
+    }
+
+    return {
+      [importFnOrProp]: {
+        value,
+        fragments,
+        [NODE_TYPE_HIDDEN_PROP]: BlockChildType.Text as BlockChildType.Text,
+      },
+    };
```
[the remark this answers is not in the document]
gohabereg: @copilot Fix it using keypath.set, add a tests covering that case

packages/core/src/components/BlockManager.ts
```diff
@@ -207,6 +212,150 @@ export class BlocksManager {
...
+      return;
+    }
+
+    /**
+     * Remove text in the split input (fragments will be adjusted by the model)
+     */
+    if (offset < splitInput.value.length) {
+      this.#model.removeText(this.#config.userId, blockIndex, dataKey, offset);
+    }
```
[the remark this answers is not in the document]
gohabereg: @copilot Add offset validation, throw an error if offset is less than 0 or more than input's content length. Add test to cover those cases

e11sy (approved):

packages/core/src/components/BlockManager.ts
```diff
@@ -207,6 +212,156 @@ export class BlocksManager {
...
+  public splitBlock(blockIndexOrId: number | BlockId, dataKey: DataKey, offset: number): void {
+    const blockIndex = this.#model.resolveBlockIndex(blockIndexOrId);
+
+    const block = this.#model.getBlockSerialized(blockIndex);
+    const toolName = block.name;
+
+    const tool = this.#toolsManager.blockTools.get(toolName);
+    const canSplit = tool?.options.canSplit === true;
+
+    const blockTextContent = Object.entries(
```
neSpecc:
```suggestion
    const blockTexts = Object.entries(
```

ilyamore88 (approved):

packages/model/src/utils/textUtils.spec.ts
```diff
@@ -0,0 +1,155 @@
...
+  };
+}
+
+describe('textUtils', () => {
+  describe('sliceFragments()', () => {
+    it('should return an empty array when given no fragments', () => {
+      expect(sliceFragments([], 0)).toEqual([]);
+    });
+
+    it('should drop fragments that end at or before the offset', () => {
```
neSpecc, line 26: why it just drops fragment instead of splitting it?

packages/model/src/utils/textUtils.spec.ts
```diff
@@ -0,0 +1,155 @@
...
+        value: 'first\n',
+        fragments: [],
+      };
+      const entries: [string, TextNodeSerialized][] = [
+        ['b', textNode('second')],
+      ];
+
+      const result = mergeTextNodes(entries, initial);
+
+      expect(result.value).toBe('first\nsecond\n');
```
neSpecc, line 112: it is not really straightforward why `mergeTextNodes` adds '\n' at the end. Maybe it should be moved to an argument like "joiner"?

packages/sdk/src/entities/BlockTool.ts
```diff
@@ -23,14 +42,18 @@ export enum BlockToolOptionKey {
   /** Inline tools enabled for blocks of this type. */
   InlineToolbar = 'inlineToolbar',
   /** Block tunes enabled for blocks of this type. */
-  Tunes = 'tunes'
+  Tunes = 'tunes',
+  /** Conversion configuration for this tool. */
+  ConversionConfig = 'conversionConfig',
+  /** Whether the block can be split into multiple blocks of its type. */
+  CanSplit = 'canSplit'
```
neSpecc: maybe `canBeSplit`?

packages/sdk/src/tools/facades/BaseToolFacade.spec.ts
```diff
@@ -161,4 +164,72 @@ describe('BaseToolFacade (via BlockToolFacade)', () => {
...
+      expect(result).toEqual({
+        text: {
+          value: 'hello',
+          fragments,
+          [NODE_TYPE_HIDDEN_PROP]: BlockChildType.Text,
+        },
+      });
+    });
+
+    it('creates nested structure when conversionConfig.import is a dot-notation key', () => {
```
neSpecc, line 210: add a test case where the `import` is a function

packages/model/src/utils/textUtils.spec.ts
gohabereg: This function slices the fragments, so it returns only content after the split point

packages/sdk/src/tools/facades/BaseToolFacade.spec.ts
gohabereg: There is one at line 177

neSpecc (approved):

PR #142, opened 2026-05-13
"Model" API

gohabereg:
- Added some API methods to be able to update the model
- Removed model dependency from AdapterPlugin & DOMAdapters

In a separate PR need to update CollaborationManager to use API as well instead of the model directly

A question for discussion
Each method would require an optional `userId` parameter for collaboration manager (as it will be dispatching updates from remove users). Maybe we'd like to have parameters as an object to reduce the number of them?

### DIFF
--- a/packages/sdk/src/api/TextAPI.ts
+++ b/packages/sdk/src/api/TextAPI.ts
@@ -1,0 +1,96 @@
+import type {
+  InlineFragment
+} from '@editorjs/model';
+
+/**
+ * Editor's TextAPI to work with text content of the document
+ */
+export interface TextAPI {
+  /**
+   * Inserts text at a given position
+   * @param text - new text to insert
+   * @param blockIndexOrId - block index or identifier
+   * @param dataKey - data key of the text node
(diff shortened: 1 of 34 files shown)


gohabereg:
As a todo, would be good to implement some safeguards for the parameters

packages/sdk/src/api/TextAPI.ts
```diff
@@ -0,0 +1,96 @@
...
+   * @param text - new text to insert
+   * @param blockIndexOrId - block index or identifier
+   * @param dataKey - data key of the text node
+   * @param start - start offset
+   */
+  insert(
+    text: string,
+    blockIndexOrId: number | string,
+    dataKey: string,
+    start?: number
```
neSpecc: I think object will be more scalable

ilyamore88 (approved):

e11syy (approved):

e11sy (approved):
conflicts remain

neSpecc (approved):

PR #145, opened 2026-05-19
Imlement simple history manager in core package

gohabereg:
Core now has simple undo redo manager for single-user use case.

Plugins can prevent default undo behaviour by cancelling UndoCore and RedoCore events. CollaborationManager prevents default behaviour to use it's own implementation instead

### DIFF
--- a/packages/collaboration-manager/src/BatchedOperation.ts
+++ b/packages/collaboration-manager/src/BatchedOperation.ts
@@ -107,7 +107,6 @@ export class BatchedOperation<T extends OperationType = OperationType> extends O
    * Checks if operation can be added to the batch
    *
    * Only text operations with the same type (Insert/Delete) on the same block and data key could be added
-   * @todo delete operations are not being batched properly
    * @param op - operation to check
    */
   public canAdd(op: Operation): boolean {
@@ -143,6 +142,20 @@ export class BatchedOperation<T extends OperationType = OperationType> extends O
       return false;
     }
 
-    return op.index.textRange![0] === lastOp.index.textRange![1] + 1;
+    if (op.type === OperationType.Insert) {
+      /**
+       * For Insert operations, each character is appended sequentially:
+       * [0,0], [1,1], [2,2] ...
+       */
+      return op.index.textRange![0] === lastOp.index.textRange![1] + 1;
+    }
+
+    /**
+     * For Delete operations two consecutive patterns are allowed:
+     * - Backspace: each deletion decrements the position: [3,3], [2,2], [1,1] ...
+     * - Forward delete: the position stays the same after each deletion: [0,0], [0,0] ...
+     */
+    return op.index.textRange![0] === lastOp.index.textRange![0] - 1 ||
+      op.index.textRange![0] === lastOp.index.textRange![0];
   }
 }
--- a/packages/collaboration-manager/src/BatchedOperation.ts
+++ b/packages/collaboration-manager/src/BatchedOperation.ts
@@ -107,6 +107,7 @@ export class BatchedOperation<T extends OperationType = OperationType> extends O
    * Checks if operation can be added to the batch
    *
    * Only text operations with the same type (Insert/Delete) on the same block and data key could be added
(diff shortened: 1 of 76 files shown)


packages/collaboration-manager/test/mocks/createManager.ts
```diff
@@ -37,16 +37,18 @@ function createMockDocumentAPI(model: EditorJSModel): DocumentAPI {
...
-  eventBus: EventBus; } {
+export function createManager(config: CoreConfigValidated, model: EditorJSModel): {
+  manager: CollaborationManager;
+  eventBus: EventBus;
+} {
   const eventBus = new EventBus();
 
   const api: EditorAPI = {
     document: createMockDocumentAPI(model),
     // eslint-disable-next-line @typescript-eslint/no-explicit-any
```
e11sy: seems like any is gone, strange that eslint does not complain about redundant ignore

packages/core/src/components/UndoRedoManager.ts
```diff
@@ -0,0 +1,344 @@
+import 'reflect-metadata';
```
e11sy, line 1: maybe we could import it once in the core src/index?

packages/core/src/components/UndoRedoManager.ts
```diff
@@ -0,0 +1,344 @@
...
+      return newRange[0] === lastRange[1] + lastPayload.length;
+    } else {
+      return (newRange[1] === lastRange[0]) || (newRange[0] === lastRange[0]);
+    }
+  }
+
+  /**
+   * Moves the batch to the undo stack and cancels the debounce timer.
+   */
+  #flushBatch(): void {
```
e11sy: i remember we discussed the same naming CollaborationManager

```suggestion
  #putBatchToUndo(): void {
```

packages/model/src/EventBus/types/EventPayloadBase.ts
e11sy, line 1: i suppose that the file should be renamed then

packages/core/src/components/UndoRedoManager.ts
gohabereg: Best practice is to include it in each file. During the build it just needs to be optimised to be included in the bundle only once

packages/core/src/components/UndoRedoManager.ts
```diff
@@ -0,0 +1,344 @@
...
+
+  /**
+   * Stack of events or batches that can be redone.
+   */
+  #redoStack: EventPayloadBase<EventAction>[][] = [];
+
+  /**
+   * Array that stores batched events
+   */
+  #batch: EventPayloadBase<EventAction>[] | null = null;
```
ilyamore88, line 64: Docs are not clear to me. Maybe highlight that this is a temporary stack with events for batching, not batched events.

e11sy (approved):

packages/collaboration-manager/src/BatchedOperation.ts
```diff
@@ -107,6 +107,7 @@ export class BatchedOperation<T extends OperationType = OperationType> extends O
    * Checks if operation can be added to the batch
    *
    * Only text operations with the same type (Insert/Delete) on the same block and data key could be added
+   * @todo delete operations are not being batched properly
```
neSpecc: @copilot implement an fix for this todo

packages/core/src/components/UndoRedoManager.ts
```diff
@@ -0,0 +1,331 @@
...
+    this.#isApplying = true;
+
+    try {
+      events.forEach(e => this.#apply(e));
+    } finally {
+      this.#isApplying = false;
+    }
+
+    this.#redoStack.push(events.map(this.#inverse).reverse());
+  }
```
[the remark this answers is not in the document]
neSpecc: is it actual?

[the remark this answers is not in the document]
gohabereg: No, fixed with arrow function

neSpecc (approved):

PR #148, opened 2026-05-24
Implement block conversion — export side and convert API

Reversean:
## Summary

- Add `exportTextContent()` to `BlockToolFacade`. Supports a function, a string keypath, or a dot-notation keypath via `conversionConfig.export`.
- Add `convertBlock()` to `BlocksManager`: exports text from the source block using its conversion config, imports it into the target tool type, and replaces the block at the same index.

### DIFF
--- a/packages/core/src/api/BlocksAPI.ts
+++ b/packages/core/src/api/BlocksAPI.ts
@@ -208,4 +208,24 @@ export class BlocksAPI implements BlocksApiInterface {
   public split({ block, key, offset, userId }: Parameters<BlocksApiInterface['split']>[0]): void {
     this.#blocksManager.splitBlock(block as BlockIndexOrId, createDataKey(key), offset, userId);
   }
+
+  /**
+   * Converts a block to a new type
+   * @param params - conversion parameters
+   * @param params.block - index or id of the block to convert
+   * @param params.key - data key of the text input at which to convert
+   * @param params.newType - block tool name to convert to
+   * @param [params.dataOverrides] - optional data overrides for the new block. Merged shallowly:
+   *   object-valued keys are replaced wholesale, not deep-merged.
+   * @param [params.userId] - user id to attribute the change to
+   */
+  public convert({
+    block,
+    key,
+    newType,
+    dataOverrides,
+    userId = this.#config.userId,
+  }: Parameters<BlocksApiInterface['convert']>[0]): void {
+    this.#blocksManager.convertBlock(block as BlockIndexOrId, createDataKey(key), newType, userId, dataOverrides);
+  }
 }
--- a/packages/core/src/components/BlockManager.spec.ts
+++ b/packages/core/src/components/BlockManager.spec.ts
@@ -680,4 +680,172 @@ describe('BlocksManager (unit, mocked deps)', () => {
       );
     });
   });
+
+  describe('.convertBlock()', () => {
+    beforeEach(() => {
+      model.resolveBlockIndex = jest.fn(() => 0);
+      model.getBlockSerialized = jest.fn(() => ({
+        name: 'header',
+        id: 'b1',
+        data: {
+          text: {
+            value: 'Hello',
+            fragments: []
+          }
+        }
+      }));
+      model.getBlockTextContent = jest.fn(() => ({
+        text: {
+          value: 'Hello',
+          fragments: []
+        }
+      }));
+    });
+
+    it('should export text from source, imports into target, and replaces block at the same index', () => {
+      const sourceTool = { exportTextContent: jest.fn(() => 'Hello') };
+      const targetTool = {
+        importTextContent: jest.fn(() => ({
+          text: {
+            value: 'Hello',
+            fragments: []
+          }
+        }))
+      };
+
+      // @ts-expect-error — mock
+      toolsManager.blockTools.get = jest.fn((name: string) =>
+        name === 'header' ? sourceTool : targetTool
+      );
+
+      blocksManager.convertBlock(0, 'text' as DataKey, 'paragraph');
+
+      expect(sourceTool.exportTextContent).toHaveBeenCalledWith({
+        text: {
+          value: 'Hello',
+          fragments: []
+        }
+      });
+      expect(targetTool.importTextContent).toHaveBeenCalledWith('Hello', []);
+      expect(model.removeBlock).toHaveBeenCalledWith(USER_ID, 0);
+      expect(model.addBlock).toHaveBeenCalledWith(
+        USER_ID,
+        {
+          name: 'paragraph',
+          data: {
+            text: {
+              value: 'Hello',
+              fragments: []
+            }
+          }
+        },
+        0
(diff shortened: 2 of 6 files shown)


packages/core/src/api/BlocksAPI.ts
```diff
@@ -124,4 +124,19 @@ export class BlocksAPI implements BlocksApiInterface {
...
+   * Converts a block to a new type
+   * @param id - id of the block to convert
+   * @param newType - block tool name to convert to
+   * @param dataOverrides - optional data overrides for the new block @todo implement
+   */
+  public convert(
+    id: string,
+    newType: string,
+    // eslint-disable-next-line @typescript-eslint/no-unused-vars
+    dataOverrides?: BlockToolData
```
gohabereg: Better to leave a to-do if you're not going to implement it in this PR

packages/core/src/api/BlocksAPI.ts
```diff
@@ -124,4 +124,19 @@ export class BlocksAPI implements BlocksApiInterface {
...
   }
+
+  /**
+   * Converts a block to a new type
+   * @param id - id of the block to convert
+   * @param newType - block tool name to convert to
+   * @param dataOverrides - optional data overrides for the new block @todo implement
+   */
+  public convert(
+    id: string,
```
gohabereg: Please use id or index

Also, it would be more scalable to use a single object with parameters and add an optional useId property (see my PR that adds user ID to API methods)

packages/core/src/api/BlocksAPI.ts
Reversean: Now we're using this parameters to override data.

packages/core/src/api/BlocksAPI.ts
Reversean: > Please use id or index

Done, now we accepting both

> Also, it would be more scalable to use a single object with parameters and add an optional useId property (see my PR that adds user ID to API methods)

Good advise, done.

gohabereg (approved):

packages/core/src/components/BlockManager.spec.ts
```diff
@@ -328,6 +328,99 @@ describe('BlocksManager (unit, mocked deps)', () => {
...
+      model.resolveBlockIndex = jest.fn(() => 0);
+      model.getBlockSerialized = jest.fn(() => ({
+        name: 'header',
+        id: 'b1',
+        data: { text: { value: 'Hello',
+          fragments: [] } },
+      }));
+    });
+
+    it('exports text from source, imports into target, and replaces block at the same index', () => {
```
neSpecc: it is a good practice to name a test case  starting with "should ...", so the full test will be in a declarative way ("it should ...")

```suggestion
    it('should export text from source, imports into target, and replaces block at the same index', () => {
```

packages/core/src/components/BlockManager.spec.ts
```diff
@@ -328,6 +328,99 @@ describe('BlocksManager (unit, mocked deps)', () => {
...
+      const sourceTool = { exportTextContent: jest.fn(() => 'Hello') };
+      const targetTool = { importTextContent: jest.fn(() => ({ text: { value: 'Hello',
+        fragments: [] } })) };
```
neSpecc: bad line breaks, please format objects in a way where each object property stays on the separate line: 

```ts
{
  text: {
    value: 'Hello',
    fragments: []
  }
}
```

packages/core/src/components/BlockManager.spec.ts
```diff
@@ -328,6 +328,99 @@ describe('BlocksManager (unit, mocked deps)', () => {
...
+
+      expect(sourceTool.exportTextContent).toHaveBeenCalledWith({ text: { value: 'Hello',
+        fragments: [] } });
+      expect(targetTool.importTextContent).toHaveBeenCalledWith('Hello', []);
+      expect(model.removeBlock).toHaveBeenCalledWith(USER_ID, 0);
+      expect(model.addBlock).toHaveBeenCalledWith(
+        USER_ID,
+        { name: 'paragraph',
+          data: { text: { value: 'Hello',
+            fragments: [] } } },
```
neSpecc: bad formatting

packages/core/src/components/BlockManager.spec.ts
Reversean: Done.

packages/core/src/components/BlockManager.spec.ts
Reversean: Fixed.

packages/core/src/components/BlockManager.spec.ts
Reversean: Fixed.

packages/core/src/components/BlockManager.ts
```diff
@@ -362,6 +362,34 @@ export class BlocksManager {
...
+  public convertBlock(blockId: string | number, newType: string, userId: string | number = this.#config.userId, dataOverrides?: BlockToolData): void {
+    const blockIndex = this.#model.resolveBlockIndex(blockId as BlockId);
+
+    const block = this.#model.getBlockSerialized(blockIndex);
+
+    const sourceTool = this.#toolsManager.blockTools.get(block.name)!;
+    const targetTool = this.#toolsManager.blockTools.get(newType)!;
+
+    const text = sourceTool.exportTextContent(block.data);
+    const newData = targetTool.importTextContent(text, []);
```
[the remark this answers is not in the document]
Reversean: Done.

packages/core/src/components/BlockManager.ts
```diff
@@ -362,6 +362,34 @@ export class BlocksManager {
...
+   * @param [userId] - user id to attribute the change to
+   * @param [dataOverrides] - optional data fields to merge on top of the converted data
+   */
+  public convertBlock(blockId: string | number, newType: string, userId: string | number = this.#config.userId, dataOverrides?: BlockToolData): void {
+    const blockIndex = this.#model.resolveBlockIndex(blockId as BlockId);
+
+    const block = this.#model.getBlockSerialized(blockIndex);
+
+    const sourceTool = this.#toolsManager.blockTools.get(block.name)!;
+    const targetTool = this.#toolsManager.blockTools.get(newType)!;
```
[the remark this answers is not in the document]
Reversean: Done.

packages/core/src/components/BlockManager.ts
```diff
@@ -362,6 +362,34 @@ export class BlocksManager {
...
+
+    const text = sourceTool.exportTextContent(block.data);
+    const newData = targetTool.importTextContent(text, []);
+    const finalData = dataOverrides !== undefined
+      ? { ...newData,
+          ...dataOverrides }
+      : newData;
```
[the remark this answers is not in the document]
Reversean: Added doc for all parameter entry along all `convertBlock()` pipeline.

neSpecc (approved):

PR #153, opened 2026-06-07
InlineToolbar implementation

gohabereg:
Still lots to do — left several @todo across the code.
Need to update the test cases for changed functions

1. InlineToolbar now renders InlinePopover
2. Inline Tools return popover items via `getToolbarConfig` method

### DIFF
--- a/packages/core/src/tools/internal/inline-tools/bold/index.ts
+++ b/packages/core/src/tools/internal/inline-tools/bold/index.ts
@@ -1,9 +1,10 @@
-import type { ToolFormattingOptions, InlineTool, InlineToolConstructor } from '@editorjs/sdk';
+import type { ToolFormattingOptions, InlineTool, InlineToolConstructor, MenuConfig } from '@editorjs/sdk';
 import { ToolType } from '@editorjs/sdk';
 import type { InlineFragment, TextRange } from '@editorjs/model';
 import { FormattingAction } from '@editorjs/model';
 import { IntersectType } from '@editorjs/model';
 import { make } from '@editorjs/dom';
+import { IconBold } from '@codexteam/icons';
 
 /**
  * Bold Tool
@@ -43,6 +44,15 @@ export class BoldInlineTool implements InlineTool {
    * If two fragment intersect, they should be merged
    */
   public intersectType: IntersectType = IntersectType.Extend;
+
+  /**
+   * Returns inline toolbar configuration for the tool
+   */
+  public getToolbarConfig(): MenuConfig {
+    return {
+      icon: IconBold,
+    };
+  }
 
   /**
    * Renders wrapper for tool without actual content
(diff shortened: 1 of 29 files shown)


packages/core/src/tools/internal/inline-tools/link/index.ts
```diff
@@ -47,6 +61,89 @@ export class LinkInlineTool implements InlineTool {
...
+              this.#api.selection.applyInlineTool({
+                tool: LinkInlineTool.name,
+                data: { href: linkInput.value },
+                caretIndex: caretIndex!,
+                /** @todo Replace link instead of applying the formatting again. Needs to be implemented in the model */
+                action: isActive ? FormattingAction.None : FormattingAction.Format,
+                keepSelection: false,
+              });
+            }
+          });
```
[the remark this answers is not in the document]
gohabereg: Input is destroyed every time popover is closed

packages/core/src/tools/internal/inline-tools/bold/index.ts
```diff
@@ -44,6 +45,15 @@ export class BoldInlineTool implements InlineTool {
    */
   public intersectType: IntersectType = IntersectType.Extend;
 
+  /**
+   * Returns inline toolbar configuration for the tool
+   */
+  public getToolbarConfig(): MenuConfig {
```
neSpecc, line 51: maybe it should be a static getter? like 
```ts
static get toolbox() {}
``` 

or 

```ts
static get toolbar() {}
```

neSpecc: or not static

packages/core/src/tools/internal/inline-tools/link/index.ts
```diff
@@ -47,6 +61,89 @@ export class LinkInlineTool implements InlineTool {
...
+        isFlippable: false,
+        items: [{
+          type: PopoverItemType.Html,
+          element: linkInput,
+        }],
+        isOpen: isActive,
+        onOpen: (close: (parent?: boolean) => void): void => {
+          const caretIndex = this.#api.selection.caretIndex;
+
+          linkInput.addEventListener('keydown', (event: KeyboardEvent) => {
```
neSpecc, line 119: when event listener will be removed?

packages/core/src/tools/internal/inline-tools/bold/index.ts
gohabereg: If it's going to be a getter we won't be able to pass parameters (link tool uses them). Definitely shouldn't be static to have access to isActive method

packages/core/src/tools/internal/inline-tools/link/index.ts
gohabereg: When popover is destroyed, input element is destroyed as well

neSpecc (approved):

e11sy (approved):

PR #156, opened 2026-06-20
Extract common types from model to new package

gohabereg:
Added model-types packages with the types that might be used by external developers.

model-types package is intended to be used only within model and sdk packages. Tools/Plugins should only import sdk package, mode package only should be used by internal packages such as core and ot-server. There is yarn workspace constraint now to check this on precommit (yarn.config.cjs file)

## Copilot instructions
1. Make sure there's no duplicate definitions (all definitions are migrated to model-types package)
2. model-types package is used only within model and sdk packages
3. model-types entities are not re-exported from the model, only from sdk
4. There's no changes to business logic, only imports updates


gohabereg:
If anyone wanna re-run review, I've hit a limit

neSpecc (approved):

packages/sdk/src/index.ts
```diff
@@ -2,3 +2,66 @@ export * from './entities/index.js';
...
+  DataNodeAddedEvent,
+  DataNodeRemovedEvent,
+  ValueModifiedEvent,
+  TextAddedEvent,
+  TextRemovedEvent,
+  TextFormattedEvent,
+  TextUnformattedEvent,
+  CaretManagerCaretUpdatedEvent,
+  BlockAddedEvent,
+  BlockRemovedEvent
```
Reversean: Should we re-export `CaretManagerCaretAddedEvent`, `CaretManagerCaretRemovedEvent`, `PropertyModifiedEvent`, `TuneModifiedEvent` here?

According on `events-catalog.mmd`, I assume that yes.

packages/collaboration-manager/tsconfig.build.json
```diff
@@ -12,10 +12,10 @@
   ],
   "references": [
     {
-      "path": "../model/tsconfig.build.json"
+      "path": "../sdk/tsconfig.build.json"
     },
     {
-      "path": "../sdk/tsconfig.build.json"
+      "path": "../model/tsconfig.build.json"
```
Reversean: Since `@editorjs/model` was moved to `devDependencies`, should `../model/tsconfig.build.json` be here?

`tsconfig.json` doesn't have it.

https://github.com/editor-js/document-model/blob/33b279c0f54f416eb1b2f6a9b5eb2af928650b0a/packages/collaboration-manager/tsconfig.json#L17-L23)

packages/model/src/entities/EditorDocument/index.ts
```diff
@@ -1,40 +1,40 @@
...
-  TextNodeEvents,
-  ValueNodeEvents
-} from '../../EventBus/types/EventMap.js';
 import {
   BlockAddedEvent,
   BlockRemovedEvent,
-  PropertyModifiedEvent, type TextFormattedEventData, type TextUnformattedEventData
-} from '../../EventBus/events/index.js';
+  PropertyModifiedEvent,
+  TuneModifiedEvent
```
Reversean: Seems like unused import.


```suggestion
  PropertyModifiedEvent
```

Reversean (changes requested):

packages/collaboration-manager/tsconfig.build.json
gohabereg: Needs to be there because model is used in tests. Maybe we'll refactor tests later on, but CI is failing without a reference for now

Reversean (approved):

Issue #150, opened 2026-05-28
Block Tunes


gohabereg:
Block Tunes is type of Tool that can enhance a Block within the editor. It's either could just a functional feature (like moving/removing blocks) or some UI enhancement (e.g. Block background, Spoiler for a Block, etc.)

Tunes are integrated in the UI within the Block Settings

<img width="246" height="247" alt="Image" src="https://github.com/user-attachments/assets/7912bb0f-de33-4ce2-ac2c-f5da3e0e12e9" />

In v2 the functional side of the Tune works via EditorAPI. The UI side works by wrapping the Block content with an element that Tune has control of.
For example:
```html
<div class="background-tune"> <!-- Tune element -->
  <div class="block-content">...</div>
</div>
```
V2 BlockTune and it's constructable interfaces:
```typescript
export interface BlockTune {
  /**
   * Returns BlockTune's UI.
   * Should return either MenuConfig (recommended) (@see https://editorjs.io/menu-config/)
   * or HTMLElement (UI consitency is not guaranteed)
   */
  render(): HTMLElement | MenuConfig;

  /**
   * Method called on Tool render. Pass Tool content as an argument.
   *
   * You can wrap Tool's content with any wrapper you want to provide Tune's UI
   *
   * @param {HTMLElement} pluginsContent — Tool's content wrapper
   */
  wrap?(pluginsContent: HTMLElement): HTMLElement;

  /**
   * Called on Tool's saving. Should return any data Tune needs to save
   *
   * @return {BlockTuneData}
   */
  save?(): BlockTuneData;
}

export interface BlockTuneConstructable {

  /**
   * Flag show Tool is Block Tune
   */
  isTune: boolean;

  /**
   * Tune's sanitize configuration
   */
  sanitize?: SanitizerConfig;

  /**
   * @constructor
   *
   * @param config - Block Tune config
   */
  new(config: {
    api: API,
    config?: ToolConfig,
    block: BlockAPI,
    data: BlockTuneData,
  }): BlockTune;

  /**
   * Tune`s prepare method. Can be async
   * @param data
   */
  prepare?(): Promise<void> | void;

  /**
   * Tune`s reset method to clean up anything set by prepare. Can be async
   */
  reset?(): void | Promise<void>;
}
```

For V3 there are going to be changes
| Method | Comment |
| ------- | ---------- |
| save() | not required anymore as data should be saved within the Model |
| render() | should be completely replaced by MenuConfig (maybe even could be a static property |
[...]

PR #157, opened 2026-06-24
feat: implement Block Tunes

bettysteger:
https://github.com/editor-js/document-model/issues/150


packages/ui/src/BlockTunes/BlockTunesUI.ts
```diff
@@ -0,0 +1,117 @@
...
+      this.#popover.addItem({
+        name,
+        title: tune.title,
+        icon: tune.icon,
+        closeOnActivate: true,
+        isDisabled: tune.isDisabled?.(),
+        onActivate: () => {
+          tune.activate?.();
+        },
+      });
```
gohabereg, line 101: I think a Tune (or a Plugin if we go with the approach I suggest) can return the configuration (or partial configuration) the same way InlineTool does.

A question however — is it better to call some instance method onActivate or instance would just provide onActivate in the config.

packages/core/src/tunes/internal/delete-block/index.ts
```diff
@@ -0,0 +1,40 @@
...
+   */
+  constructor({ api, blockId }: BlockTuneConstructorOptions) {
+    this.#api = api;
+    this.#blockId = blockId;
+  }
+
+  /**
+   * Deletes the block this tune is attached to
+   */
+  public activate(): void {
```
gohabereg, line 33: In v2 this tune has confirmation state as well (could be done via Popover)

gohabereg (commented):
Great work! Going through the PR started my train of thought on the Tunes overall. I invite you all for a discussion @bettysteger @neSpecc @e11sy 

Below is just my stream of consciousness, hope it makes sense

> TL;DR
> I suggest to remove BlockTune as an entity and just use Plugins. That could be possible with 3 core changes:
> 1. UI Plugin would have an API
> 1.1. API would have a capability to add a button to the Settings popover
> 1.2. API would also have a capability to change a Block appearance via a wrapper (the same way it's done in v2 now)
> 2. Instead of `tunes` property in the BlockNode, we'd have `plugins`. Any Plugin would be able to save data to the Model and retrieve it.
> 3. "Tune" becomes a Plugin that utilises UI Plugin APIs and of course Core APIs to change the Model
> 
> The "Tune" term than becomes just a UI term, not an entity. Therefore there's no need to have a separate BlockTuneAdapter.
 
The purpose of the Adapter is to connect the Tune instance to the Model. In the PR there's no such connection — Tune updates the model through the Core API rather than Adapter and Adapter doesn't handle Model updates (in this PR, because there's no need for internal Tunes).

Comparing TuneAdapter to FormattingAdapter, CaretAdapter and BlockToolAdapters: these 3 adapters are converting user input (typing, selection, applying inline formatting) to the Model data. It could be done without Adapters, Tools could just work with the DOM directly and update the Model via Core API. But Adapters remove boilerplate and handle complex DOM operations, so a Tool developer doesn't need to. With the Tunes there is no complex DOM operations or converting complex user input into the Model data. Ultimately, Tune is just a UI layer for the Core API calls + some of them might carry some data (e.g. Anchors Tune, or Footnotes Tune).

That kinda makes a Tune just a Plugin with the ability to add a button — maybe that's the way to go? There are use-cases when we expect the Tune to change editor's zone appearance. In v2 it's done via a Block wrapper — maybe we just add this capability to UI Plugin instead of having a separate "Tune" entity.

So the final design would look like:
1. UI being a Plugin would provide capabilities to change Block's appearance with a wrapper
2. Any Plugin can interact with another Plugin via it's public API (see a ticket on the project board to expose Plugins API)
3. Any Plugin can add a button into the Settings section of the UI
4. Tunes become just Plugins interacting with UI plugin.
5. In the Model structure instead of "tunes" property in the BlockNode we introduce something like "plugins" or "pluginsData" and add a capability for each Plugin to change that data through the Core API.

packages/core/src/components/BlockRenderer.spec.ts
```diff
@@ -84,6 +85,9 @@ describe('BlockRenderer (unit, mocked deps)', () => {
   const adapter: EditorJSAdapterPlugin = {
     createBlockToolAdapter: jest.fn(() => ({})),
     destroyBlockToolAdapter: jest.fn(),
+    createBlockTuneAdapter: jest.fn(() => ({})),
+    getBlockTuneAdapter: jest.fn(() => undefined),
+    destroyBlockTuneAdapters: jest.fn(),
```
Reversean: `toolsManager.blockTunes` in the mock is always an empty `Map()`, so the new "create a `BlockTuneAdapter` per registered tune when a block is created" loop in `BlockRenderer.ts` never actually executes in any test here, and there's no assertion that `destroyBlockTuneAdapters` is called on block removal (only `destroyBlockToolAdapter` is checked). Could use a case with a non-empty `blockTunes` map to cover both creation and teardown.

PR #158, opened 2026-07-02
refactor(paragraph): extract paragraph block tool into standalone package

Reversean:
## Summary
- Move the built-in Paragraph block tool out of `@editorjs/core` into its own publishable package, `@editorjs/paragraph`, versioned independently like `sdk`, `model`, and `dom-adapters`.
- Pure relocation — the tool's source has no relative imports into core internals, so behavior is unchanged. `core` now depends on the new package as a workspace dependency.


gohabereg (approved):

packages/tools/paragraph/package.json
```diff
@@ -0,0 +1,30 @@
+{
+  "name": "@editorjs/paragraph",
+  "version": "0.0.0",
```
neSpecc: we already have this package https://github.com/editor-js/paragraph/blob/master/package.json

I'd suggest to create a 3.0.0 version

Reversean: Done.

neSpecc (approved):

PR #161, opened 2026-07-02
refactor(link): extract link inline tool into standalone package

Reversean:
## Summary
- Move the built-in Link inline tool out of `@editorjs/core` into its own publishable package, `@editorjs/link`, completing the split of all four built-in tools started with Paragraph.
- Pure relocation — the tool's source has no relative imports into core internals, so behavior is unchanged. `core` now depends on the new package as a workspace dependency, and no longer needs `@editorjs/dom` or `@codexteam/icons` directly (only the extracted tool packages use them now).

Stacked on #160 (italic).


README.md
```diff
@@ -25,6 +25,7 @@ A model-driven, collaboration-ready Editor.js engine split into focused packages
 | [`@editorjs/paragraph`](packages/tools/paragraph) | Built-in Paragraph block tool |
 | [`@editorjs/bold`](packages/tools/bold) | Built-in Bold inline tool |
 | [`@editorjs/italic`](packages/tools/italic) | Built-in Italic inline tool |
+| [`@editorjs/link`](packages/tools/link) | Built-in Link inline tool |
```
gohabereg: Probably better to add a separate section for tools/plugins

gohabereg (approved):

README.md
Reversean: Ok, moved all tool entries (including #158 #159 #160) in additional section.

Please, check this out.

packages/tools/link/package.json
```diff
@@ -0,0 +1,29 @@
+{
+  "name": "@editorjs/link",
```
neSpecc: we already have this package, and it is not inline-link, but embedded link block

https://github.com/editor-js/link/blob/master/package.json

neSpecc (changes requested):

packages/tools/link/package.json
Reversean: Renamed package name into `@editorjs/inline-link`.

neSpecc (approved):

PR #162, opened 2026-07-02
feat: add @editorjs/header package

Reversean:
Header lives in its own repo ([editor-js/header](https://github.com/editor-js/header)). `header` also needs `model`, `sdk`, and `dom-adapters` as deps while it's being built against them, so it can be edited against the live, unpublished code in this workspace without a publish-and-bump cycle for every change to those packages.

This adds packages/tools/header as a submodule. The .gitmodules url is still a local filesystem path, not the GitHub repo, so that needs updating separately before this lands. 

The 8 workflow files that run the root setup action now pass `submodules: true` to `actions/checkout`.

### DIFF
--- a/packages/core/src/components/BlockManager.ts
+++ b/packages/core/src/components/BlockManager.ts
@@ -145,7 +145,7 @@ export class BlocksManager {
     }
 
     this.#model.addBlock(userId, {
-      ...data,
+      data,
       id,
       name: type,
     }, newIndex);
--- a/.github/actions/base-coverage/action.yml
+++ b/.github/actions/base-coverage/action.yml
@@ -11,6 +11,10 @@ inputs:
 runs:
   using: "composite"
   steps:
+    - name: Deinit submodules before comparing against base ref
+      shell: bash
+      run: git submodule deinit --all -f
+
     - name: Checkout base ref
       uses: actions/checkout@v6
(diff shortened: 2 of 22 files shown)


packages/core/src/components/BlockManager.ts
```diff
@@ -144,7 +145,7 @@ export class BlocksManager {
     }
 
     this.#model.addBlock(userId, {
-      ...data,
+      data,
       id,
```
gohabereg, line 149: Please explain the change

gohabereg (commented):
We have header tool in a separate repo, and there's already a PR for v3 — https://github.com/editor-js/header/pulls

I think it would be better to add it here as a submodule, so the separate repo stays

packages/core/src/components/BlockManager.ts
Reversean: [`BlockNode`](https://github.com/editor-js/document-model/blob/4987d0a2c8a1ab45d80f3355c3f23424943bb160/packages/model/src/entities/BlockNode/index.ts#L74) object created during `addBlock` execution expects field `data`, so i thought we shouldn't flatten it here, instead we should provide `data` object with additional block preferences such as header level.

gohabereg (approved):

PR #163, opened 2026-07-02
feat(playground): register Header tool

Reversean:
Registers @editorjs/header in the playground.

Getting Header to actually render surfaced two bugs, both fixed here. 

- `BlockManager.insert` spread a block's data onto the block object's own top-level properties instead of nesting it under data, so every tool lost its data on insert. 
- `ToolboxUI.addTool` read `tool.options.toolbox` and added a single popover entry per tool; it now iterates `tool.toolbox`, since a tool can define several entries (Header has one per heading level) and only the first ever showed up.

### DIFF
--- a/packages/core/src/components/BlockManager.spec.ts
+++ b/packages/core/src/components/BlockManager.spec.ts
@@ -143,6 +143,22 @@ describe('BlocksManager (unit, mocked deps)', () => {
       );
     });
 
+    it('should nest data under a data property instead of flattening it onto the block', () => {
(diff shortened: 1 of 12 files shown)


gohabereg (commented):
Let's add `@editorjs/editorjs` pacakge inside the monorepo. Mark it private for now.

This package should import Core and all the Tools and register them. Export the core with registered tools.

Playground should use import from editorjs/editorjs package instead of core.

Registrations inside the core could be removed

PR #168, opened 2026-07-17
Init openspec & generate specs on existing codebase

gohabereg:


openspec/config.yaml
```diff
@@ -0,0 +1,20 @@
...
+
+# Per-artifact rules (optional)
+# Add custom rules for specific artifacts.
+# Example:
+#   rules:
+#     proposal:
+#       - Keep proposals under 500 words
+#       - Always include a "Non-goals" section
+#     tasks:
+#       - Break tasks into chunks of max 2 hours
```
Reversean: Maybe we could fill this config? Something like:

```suggestion
context: |
  Tech stack: TypeScript, Yarn 4 (Berry) workspaces monorepo (packages/*, packages/tools/*).
  Domain: Editor.js document model - a collaborative rich-text editor engine (block/inline
  data tree, operational-transform based collaboration, pluggable block/inline tools).
  Key packages: model (EditorJSModel/EditorDocument/BlockNode/inline-fragments/CaretManager),
  model-types (shared event/type contracts), core, dom-adapters, ot-server (collaboration/OT
  server), sdk (public API for building tools/plugins - tools should depend on this, not on
  model directly), ui, collaboration-manager, and packages/tools/* (built-in block/inline
  tools implementing BlockTool/InlineTool from @editorjs/sdk).
  Existing architecture docs live in docs/; Mermaid diagrams in docs/diagrams/;
  consult these before writing new specs/proposals so they don't duplicate or contradict
  what's already documented there.
  Testing: Jest with co-located *.spec.ts/*.integration.spec.ts files; some packages also
  run Stryker mutation testing (test:mutations). Linting via ESLint (yarn lint).

rules:
  tasks:
    - Write the failing test first, then implement, then refactor (TDD).
    - Name test cases in should-notation, e.g. it('should return X when Y').
  proposal:
    - Cross-check against existing docs in docs/ and note if the proposal supersedes any of them.
```

.github/skills/openspec-apply-change/SKILL.md
```diff
@@ -0,0 +1,160 @@
+---
```
Reversean: This file (and its counterparts under `.opencode/skills/`, plus the second copy under `packages/.claude/skills/`) duplicates `.claude/skills/`.

Skills follow the open [Agent Skills](https://agentskills.io) standard, and all three tools already read `.claude/skills/` or `.agents/skills` directly. 

Claude Code also walks up to the repo root for skills, so the `packages/.claude/skills` copy isn't needed either. Recommend keeping just the root `.claude/skills/` and dropping the ~21 duplicate files across `.github/skills`, `.opencode/skills`, and `packages/.claude/skills`.

Reversean (changes requested):
Spec-driven development is a genuinely good direction, and I'd love to see this pattern spread to our other repos too.

Left a few line notes below on tightening things up.

I also wanted to point out that no CI step validates the specs (e.g. `openspec validate`). Is this planned for future changes?

gohabereg:
> I also wanted to point out that no CI step validates the specs (e.g. `openspec validate`). Is this planned for future changes?

I am not sure we need it at this point. For now I'd assume the spec and implementation would be done by the same person in the same branch, so the change spec would be verified and archived before it gets to the main branch or even committed to the feature branch.

Once we would be open to the community contributions we can think about that

Reversean (approved):

CONTRIBUTING.md
```diff
@@ -0,0 +1,142 @@
...
+```
+
+Some packages (e.g. `model`) also run Stryker mutation testing via `yarn test:mutations`.
+
+To try changes end-to-end, use the Vite playground:
+```bash
+cd packages/playground && yarn dev
+```
+
+The OT collaboration server can be run via Docker: `docker compose up`.
```
Reversean: Not sure it will work smoothly: it requires `WSS_PORT` env. Consider to mention about `.env`.

Reversean (changes requested):

Reversean (approved):

ilyamore88 (approved):

PR #179, opened 2026-07-21
refactor(clipboard-plugin): remove meta from editorjs clipboard payload

ilyamore88:
## Why

The `application/x-editor-js` clipboard payload carried a `meta.version` field hardcoded to `'3.0.0'`, alongside a `@todo get version info from Core`. Nothing reads it — no paste handler exists yet, and a repo-wide grep finds the MIME type only in this plugin and its spec.

Shipping a placeholder version in a public clipboard format is worse than shipping nothing: a future paste handler that trusted the field would branch on a value that never reflected reality. Removing it now, before a consumer exists, costs no migration.

## What changed

- `application/x-editor-js` payload is now `{ blocks }` — `meta` and its `version` are gone.
- Deleted the module-private `Meta` interface and the `meta` member of `ClipboardEditorJSObject`.
- `#createClipboardObject` returns `{ blocks }`; the `@todo` about sourcing a version from Core is dropped.
- Spec (`openspec/specs/clipboard-plugin/spec.md`) updated, plus a new scenario pinning that `blocks` is the payload's only key.

The object wrapper is kept rather than serializing the bare array, so reintroducing a sibling key later stays additive for readers.

## Breaking change

The `application/x-editor-js` clipboard payload no longer includes `meta`.

**No exported TypeScript API changes** — `ClipboardEditorJSObject` and `Meta` were module-private, so no `.d.ts` surface moves. The break is on the wire format only, and the value it carried was a placeholder, not a real version.

## Verification

- 12/12 tests pass (2 written failing first, per the repo's TDD rule)
- `lint:ci` clean at `--max-warnings 0`
- `tsc --build` clean
- Stryker: 100% mutation score (break threshold 75), 0 survivors

## Notes

Planned via OpenSpec; artifacts archived under `openspec/changes/archive/2026-07-21-remove-clipboard-payload-meta/`.

One thing worth a look: `openspec archive` rewrote `openspec/specs/clipboard-plugin/spec.md` and, as collateral, stripped blank lines around `## Requirements` and dropped the trailing ``Implemented in `src/index.ts`...`` footer. That footer is a convention in every spec file under `openspec/specs/`, so I restored it — the spec diff here reflects only the intended payload change. May be worth filing against the archive tool separately.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


gohabereg (approved):

packages/plugins/clipboard-plugin/src/index.spec.ts
```diff
@@ -224,11 +224,20 @@ describe('ClipboardPlugin', () => {
               { id: 'b2',
                 type: 'header' },
             ],
-            meta: { version: '3.0.0' },
           })
         );
       });
 
+      it('should not include meta in the editorjs clipboard payload', () => {
```
Reversean: This duplicates test case `should add custom editorjs data-type to native event`: existing case will fail anyway if `meta` appears in object. Better drop it.

Also since we use agents+SDD can we clarify for agents to not duplicate test cases just to check different effects of same action with same pre-conditions?

Reversean (changes requested):

packages/plugins/clipboard-plugin/src/index.spec.ts
ilyamore88: Agreed — the exact-string assertion in `should add custom editorjs data-type to native event` already fails if `meta` reappears, so this case added no coverage. Removed in 36a75c0; mutation score stays at 100% (0 survivors), confirming it was redundant.

packages/plugins/clipboard-plugin/src/index.spec.ts
```diff
@@ -224,11 +224,20 @@ describe('ClipboardPlugin', () => {
...
+        new ClipboardPlugin(pluginParamsMock);
+
+        const { setData } = dispatchCopyEvent();
+
+        const payload = setData.mock.calls.find(([type]) => type === 'application/x-editor-js')?.[1];
+
+        expect(Object.keys(JSON.parse(payload as string) as object)).toEqual(['blocks']);
```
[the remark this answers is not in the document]
ilyamore88: This applied to the `should not include meta...` test, which has been removed in 36a75c0 as redundant with the exact-payload assertion — so the unguarded `JSON.parse` is no longer present. The remaining payload test uses `toHaveBeenCalledWith` on the full serialized string and doesn't parse, so there's nothing left to guard.

Reversean (approved):

PR #180, opened 2026-07-21
feat(core): extract ShortcutsPlugin into @editorjs/shortcuts-plugin

ilyamore88:
Closes #171.

Extracts the last remaining in-tree `EditorjsPlugin` out of `packages/core` into a standalone `@editorjs/shortcuts-plugin` package, mirroring the `ClipboardPlugin` extraction (#129).

## What changed

- **New package** `packages/plugins/shortcuts-plugin` (`@editorjs/shortcuts-plugin`), scaffolded from `clipboard-plugin` (`package.json`, `tsconfig{,.build,.eslint}.json`, `eslint.config.mjs`, `jest.config.ts`, `stryker.conf.mjs`, `.gitignore`, `README.md`).
- **Source moved verbatim** — `ShortcutsPlugin.ts` → `src/index.ts` with **zero content change** (git detects a pure rename). The plugin already imported only from `@editorjs/sdk`, so no code edits were needed.
- **New test suite** `src/index.spec.ts` — 12 cases at **100% coverage** (statements/branches/functions/lines), closing the pre-existing gap (there was no `ShortcutsPlugin.spec.ts`). Covers shortcut registration from tool `options`, keydown match/dispatch, first-match-wins, the IME-composition guard, `IndexError` tolerance vs. rethrow of other errors, and `destroy()`.
- **Core wiring** — added the `workspace:^` dependency, added project references in both `tsconfig.json` and `tsconfig.build.json`, switched the import to `@editorjs/shortcuts-plugin`, and removed the now-empty `src/plugins/` directory. Registration order in `Core` is unchanged.
- **CI** — new `.github/workflows/shortcuts-plugin.yml`, same `package-check.yml`-based shape as `clipboard-plugin.yml` (`include-mutations: true`).
- **OpenSpec** — moved the "Keyboard shortcuts plugin" requirement out of `specs/core/spec.md` into a new `specs/shortcuts-plugin/spec.md`; `openspec validate` passes for all 12 specs.

## Notes / scope

- Per the request, the plugin **source is unchanged**. The two reserved-but-unused private methods (`#processBlockTool` / `#processBlockTune`) are moved as-is; they carry no statements so coverage stays at 100%. Consequently the issue's suggestion to also test the reserved block-tool/tune `shortcuts` map isn't applicable yet — it's documented as reserved in the new spec instead.

## Verification

- `yarn workspace @editorjs/shortcuts-plugin lint:ci` — clean
[...]

PR #166, opened 2026-07-15
Refactor Index class

gohabereg:
See docs for the changes description

PR #96, opened 2025-04-08
chore(ui): ui package bootstrapped

neSpecc:
- UI components moved to `@editorjs/ui` package
- Now they should be passed via `core.use()` method, where they will be instantiated
- each ui component is `EditorjsPlugin` — a class with `type` (used for ioc key) and `destroy()`. It accepts `EventBus, API, Config`
- `@editorjs/core` build config updated

PR #183, opened 2026-07-22
feat(editorjs): add batteries-included bundle package and make core headless

gohabereg:
## Why

Default tools, plugins, infra (DOM adapter, collaboration), and UI were hardcoded into `Core`'s constructor, so `@editorjs/core` couldn't be consumed as a pure engine and there was no batteries-included package a v2 user could drop in.

## What Changes

- **New package `@editorjs/editorjs`** (`packages/editorjs`): a batteries-included entry point whose default export `EditorJS` composes `Core` + `DOMAdapters` + `CollaborationManager` + default tools (paragraph, bold, italic, link) + plugins (clipboard, shortcuts) + UI. It **auto-initializes in the constructor** and exposes an **`isReady`** promise, and honors a v2-style **`config.tools`** map merged over the defaults with **override-by-name**.
- **`Core` is now headless** (**BREAKING**): the constructor no longer registers any default tools/plugins/infra/UI. `initialize()` **fails loudly** — it throws when no rendering adapter is registered or the `defaultBlock` tool is missing, and re-throws init failures instead of swallowing them with `console.error`.
- **`ShortcutsPlugin` is consumed from `@editorjs/shortcuts-plugin`** — extracted from core by #180, which landed on `main` while this PR was open and is merged in here. The bundle registers it as a default plugin; core neither ships nor registers it.
- **`packages/ui/index.html` fixed**: the demo page constructed `Core` with no adapter and no block tool, relying on the defaults core used to register itself, so it threw under the new fail-loud validation. It now registers `DOMAdapters` and `Paragraph` explicitly.
- **Dead deps dropped from `@editorjs/ui`**: `@editorjs/dom-adapters`, `@editorjs/core` and `editorjs-v2` were production dependencies that `packages/ui/src` never imported (dom-adapters has been unused since the package was bootstrapped in #96). Core and dom-adapters move to devDependencies for the demo page; `editorjs-v2` is dropped. Matching stale entries removed from the vite externals list and both tsconfigs' project references.
- **Dependency alias**: the published `@editorjs/editorjs` v2 dep is aliased as `editorjs-v2` (`npm:@editorjs/editorjs@…`) in `core`, `sdk`, `tools/paragraph` to free the package name. Non-behavioral; import specifiers only.
- **Playground** now consumes `@editorjs/editorjs` instead of wiring `Core` + UI by hand.

## Merged `main` (#184, #166, #185, #186, #188)

`main` is merged in as of `3b61926`. Five conflicts, three of them substantive:

- **`packages/core/src/index.ts`** — `main` had re-asserted the batteries-included constructor (`this.use(Paragraph)`, `ClipboardPlugin`, `DOMAdapters`, …) and wraps `initialize()` in a `try/catch` that logs. Kept this PR's headless + fail-loud version, and took from `main` only what was genuinely new: the `PluginRegistry` wiring, and moving `UndoRedoManager` resolution to **after** plugin/tool init. That reorder is `main`'s deliberate fix — the manager relies on `defaultPrevented` being set by plugins — and since merge-base had it before, this branch only looked like it disagreed.
- **`sdk/entities/{BaseTool,InlineTool}.ts` + `tools/paragraph`** — `main` stopped importing `ToolConfig` from the v2 package and now defines it in `BaseTool.ts`; this branch had renamed that dep to `editorjs-v2`. Took both: `ToolConfig` from the SDK, remaining v2 type imports via the alias. This is what `sdk` spec now requires anyway ("`ToolConfig` SHALL be an SDK-owned contract, not a re-exported passthrough from `@editorjs/editorjs`"). Confirmed no `ToolConfig` import resolves to the v2 package anywhere.
- **`packages/ui/package.json`** — union of both devDependency sets.

## Verification

- Full workspace **build**, **lint**, and **tests** pass post-merge. Core: 179/179. Bundle: 8/8. `openspec validate --all --strict`: 15/15.
- **New `packages/core/src/index.spec.ts`** covers the fail-loud contract, which previously had no tests at all — core had no `index.spec.ts`. It drives a real `Core` (real IoC containers and model) with stub adapter/block tool/plugin registered via `use()`, asserting both throwing paths (missing adapter, missing/mismatched `defaultBlock`) and the succeeding path. No DOM environment needed: with an empty document and stub registrations the boot sequence never reaches the holder.
- **`packages/ui/index.html` verified in-browser**: renders all three paragraphs, no console errors. The ui build output is byte-identical after the dependency cleanup (85.72 kB / 67.57 kB, 28 modules), confirming the removed externals were no-ops.
- **Playground verified end-to-end in-browser**, through the bundle: renders the configured paragraphs, typing updates the model, the inline toolbar appears on selection and its **B** button applies `<b>`.

### Keyboard shortcuts: now working (was blocked on #184)

**CMD+B works, as of the `main` merge.** It could not work on this branch before: `ShortcutsPlugin` subscribes to `ui:key-down` (`KeydownUIEvent`), but nothing dispatched that event — `packages/ui/src/Blocks/Blocks.ts` handled CMD+Z / CMD+SHIFT+Z locally and returned early for every other key. Registering the plugin was necessary but not sufficient. The producer side is the `BlocksUI` keydown dispatch from #184, now merged in at `Blocks.ts:125`.

Re-verified in the browser through the bundle, on a selection inside a paragraph:

| Action | Result |
| --- | --- |
| CMD+B | selection bolds, inline toolbar's **B** goes active |
| CMD+B again | reverts to regular weight, **B** inactive |
| CMD+I | selection italicizes, *i* goes active |

The toolbar's active state tracking the text confirms it round-tripped through the model rather than being a contenteditable side effect — `isActive` is reading back the fragment the shortcut wrote — and exercises both the Format and Unformat branches of `getFormattingOptions`. The handler table is populated from the namespaced `static options.plugins.shortcuts.shortcut` that Bold and Italic declare, which is also #184's shape.

Task 4.6 in the archived change is closed accordingly; that change now has no open tasks.

_(An earlier revision of this description claimed CMD+B was verified end-to-end, before it could possibly work. That was wrong — thanks @Reversean for pushing on it. The claim above is a fresh in-browser check made after the merge, not a restatement of the old one.)_

## Specs

Implemented via OpenSpec change `add-editorjs-bundle-package` (archived in this PR). Because the shortcuts extraction landed mid-review, the archived change was updated in place rather than superseded: proposal, design, tasks and the `editorjs-bundle` delta now reflect shortcuts being registered from its own package. Main specs updated: `core` (headless + fail-loud scenarios, now with a validation pointer to the new spec file) and the new `editorjs-bundle` capability.

Auditing all 15 specs against the merged tree turned up further drift, fixed in two commits kept deliberately separate:

- `8139d2f` **(this branch's own gap)** — `architecture` still asserted *"Default adapter is DOMAdapters"*, i.e. that `Core`'s constructor calls `this.use(DOMAdapters)`. Accurate at merge-base and still accurate on `main`, but false here, and it directly contradicted `core`'s "No adapter registered fails loudly". The bundle change wrote deltas for `core` and `editorjs-bundle` and missed that the same fact was asserted in `architecture` too.
- `43b5513` **(pre-existing on `main`, not introduced here)** — `tools` described Bold/Italic as declaring flat `shortcut: 'CMD+B'` / `'CMD+I'` properties; #184 moved them to `options.plugins.shortcuts.shortcut` without updating the spec, leaving it contradicting `shortcuts-plugin`'s own "Legacy flat shortcut key is ignored" scenario. `plugin-public-api` and `tool-plugin-options` also still carried the generated `TBD - created by archiving change add-plugin-public-api` placeholder purposes.

The second commit is unrelated to this PR's subject and is a clean standalone cherry-pick onto `main` if you'd rather it not ride along — happy to drop it.

## Follow-ups (not in this PR)

The same audit found four spec/code gaps that all trace to #184 and need code, not wording. Flagging rather than fixing, since they're outside this PR's scope:

1. **Plugin teardown is specified but unimplemented.** `core` ("Registry is cleared on teardown") and `sdk` ("Destroying a plugin") both describe `destroy()` running on teardown. `Core` has no teardown path, nothing calls plugin `destroy()`, and `PluginRegistry.unregister()` has no production caller.
2. **Duplicate/empty plugin names are only rejected when a plugin exposes a `publicApi`.** `Core#initializePlugin` guards the `registry.register()` call on `instance.publicApi !== undefined`, so two same-named plugins without one register silently — while `core` and `plugin-public-api` state the rejection unconditionally.
3. **The integrator-facing `api.plugins` surface doesn't exist.** `plugin-public-api` requires it "on the API surface available to the editor's integrator", but neither `Core` (only `use()`/`initialize()`) nor the bundle's `EditorJS` (only `isReady`) exposes one.
4. **`CollaborationManager` logs `Unknown event type DataNodeAddedEvent`** three times per page load — no case for that event, falls through to `default: console.error`. Pre-existing at merge-base and on `main`; already noted as out of scope in task 4.4.


ilyamore88 (approved):

packages/editorjs/src/index.ts
```diff
@@ -0,0 +1,103 @@
...
+    /**
+     * Infrastructure: rendering adapter and collaboration.
+     */
+    this.#core.use(DOMAdapters);
+    this.#core.use(CollaborationManager);
+
+    /**
+     * Default plugins.
+     */
+    this.#core.use(ClipboardPlugin);
```
Reversean, line 74: `ShortcutsPlugin` isn't registered here, however it was mentioned in [`openspec/specs/editorjs-bundle/spec.md`](https://github.com/editor-js/document-model/pull/183/changes#diff-379ce07a182626fdb68bc0d52158015ac0360332cc1177ec785a993235c7f0caR4) as one of default plugins.

packages/core/src/index.ts
```diff
@@ -166,32 +149,55 @@ export default class Core {
...
+    const eventBus = this.#iocContainer.get(EventBus);
+
+    eventBus.dispatchEvent(new CoreEventBase(CoreEventType.Ready, undefined));
+  }
+
+  /**
+   * Validates that the caller has registered everything the headless engine needs before booting.
+   * Throws with a clear message instead of failing deep inside module resolution.
+   */
+  #validatePreconditions(): void {
```
Reversean, line 194: Worth adding tests that constructs a real `Core`, registers a stub adapter/tool/plugin via `.use()`, and cover throw/no-throw paths?

packages/core/src/index.ts
```diff
@@ -109,15 +101,6 @@ export default class Core {
...
-
-    this.use(Paragraph);
-    this.use(BoldInlineTool);
-    this.use(ItalicInlineTool);
-    this.use(LinkInlineTool);
-    this.use(ShortcutsPlugin);
-    this.use(CollaborationManager);
-    this.use(DOMAdapters);
-    this.use(ClipboardPlugin);
   }
```
Reversean, line 106: `packages/ui/index.html` still does `new Core({...}).use(EditorjsUI).use(BlocksUI).use(InlineToolbarUI).use(ToolboxUI).initialize()` with no adapter and no block tool registered. That will now throw "No rendering adapter registered" immediately. 

Worth registering the missing things directly?

Reversean (changes requested):

packages/editorjs/src/index.ts
Reversean: Ok, read `design.md`, it explicitly defer this to a separate change. 

Still PR desc doesn't reflect that: "What Changes" lists default plugins as "clipboard, shortcuts", and "Verification" claims Cmd+B was verified end-to-end through the bundle – neither matches this diff.

Worth updating description?

Reversean (approved):

PR #188, opened 2026-07-28
feat(sdk): derive static tool options from tool config

Reversean:
## Summary

`static options` is read off the tool class before any registration exists, while
`ToolConfig` is resolved per registration. No option value can therefore depend on
config — a field meant to drive toolbox entries or a shortcut stays inert, declared
in the tool's `Config` type and silently ignored.

Two SDK changes:

- `static options` may now be a synchronous factory of the config:
  `Options | ((config: ToolConfig) => Options)`. The facade resolves it once per
  registration into a per-instance field, so nothing is written back onto the shared
  class and two `Core` instances can register one tool class with different configs.
- `ToolConfig` becomes SDK-owned instead of an `any`-defaulted re-export from
  `@editorjs/editorjs`, which silently disabled checking for tools omitting the generic.

`prepare()` is unchanged. `packages/core` is untouched — resolution lives entirely in
`BaseToolFacade`.

**BREAKING**: `ToolConfig` is imported from `@editorjs/sdk`; `static options` widens to a union.

### DIFF
--- a/packages/sdk/src/tools/facades/BaseToolFacade.ts
+++ b/packages/sdk/src/tools/facades/BaseToolFacade.ts
@@ -8,7 +8,8 @@ import { ToolType, BaseToolOptionKey } from '../../entities/index.js';
 import { type BlockTuneFacade } from './BlockTuneFacade.js';
 import type {
   BlockTool, BlockToolConstructor, InlineTool, InlineToolConstructor, BlockTuneConstructor,
-  ToolTypeToOptions, ToolStaticOptions, BlockToolOptions, InlineToolOptions, BlockTuneOptions
+  ToolTypeToOptions, ToolStaticOptions, BlockToolOptions, InlineToolOptions, BlockTuneOptions,
+  ToolOptionsFactory
 } from '../../entities/index.js';
 import type { EditorAPI } from '../../api';
 
@@ -105,6 +106,16 @@ export abstract class BaseToolFacade<Type extends ToolType = ToolType, ToolClass
   protected readonly constructable: ToolConstructable;
 
   /**
+   * Tool's static options, resolved once for this registration.
+   *
+   * When the tool declares `options` as a {@link ToolOptionsFactory}, the factory is
+   * invoked in the constructor and its result kept here — per facade instance, never
+   * written back onto the shared tool class. Two Core instances registering the same
+   * class with different configs therefore cannot observe each other's derived values.
+   */
+  protected readonly resolvedStaticOptions: ToolTypeToOptions[Type];
+
+  /**
    * Default placeholder specified in EditorJS user configuration
    */
   protected defaultPlaceholder?: string | false;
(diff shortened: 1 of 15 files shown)


openspec/changes/separate-tool-options-and-config/design.md
```diff
@@ -0,0 +1,81 @@
...
+
+**2. `options` and `config` stay two separate top-level concepts; `options.config` remains the only bridge.**
+This matches how they're actually consumed: `options` (via facade getters) is read by core/plugins before any block exists; `config` is read by the tool instance itself. Collapsing them into one `ToolSettings<Config>` bag would erase that timing distinction, which is the actual source of the bug (something that's plugin-timing data, like `toolbox`, has no way to see something that's tool-instance-timing data, like `config.levels`, unless the two are explicitly bridged).
+*Alternative considered*: a single unified settings object passed everywhere. Rejected — every plugin consumer would need to filter out user-data noise, and it doesn't resolve the "when is this available" question.
+
+**3. Config-derived options are resolved via `prepare()`'s return value, held per-facade-instance — not by writing onto `ClassName.options`.**
+`prepare()`'s signature extends to:
+```ts
+prepare?(data: { toolName: string, config: Config }):
```
gohabereg: I think going through prepare is an extra step. We can extends `options` property to be method which would accept the config

```
interface ToolConstructor {
  options: ToolOptions | (config: ToolConfig) => ToolOptions
}
```

Facade would check if options is a method and pass the config if required. So external access via Facade is unaffected

It would simplify the flow and remove new `preparedOptions` entity

Reversean: Adopted – `options` is now `ToolOptions | ((config: ToolConfig) => ToolOptions)`, resolved once in the facade. `PreparedToolOptions` and the `ToolsManager` plumbing are gone, `prepare()` is untouched.

packages/sdk/src/tools/facades/BaseToolFacade.ts
```diff
@@ -40,6 +41,28 @@ export type ToolOptions = ToolStaticOptions;
...
+ * Resolves a tool class's static options for a single registration.
+ *
+ * A tool may declare `options` either as a plain object or as a {@link ToolOptionsFactory}
+ * of its config. The factory form is invoked here — once, synchronously — with the `config`
+ * supplied to `use(Tool, { config })`, and applies its own defaults for keys the integrator
+ * omitted.
+ * @param constructable - tool class being registered
+ * @param useToolOptions - second argument of `use(Tool, options)`
+ */
+function resolveStaticOptions(constructable: ToolConstructable, useToolOptions: ToolOptions): ToolOptions {
```
gohabereg: Why not a private method?

Reversean: Moved to `BlockToolFacade` private methods.

gohabereg (approved):

neSpecc (approved):

